PR #552, opened 2021-12-08
minor style fixes for filter pane after testing in develop

g-s-k:
Story: N/A
Endpoint: TBD

### Details

Improve styling on filter pane when package list gets longer than a page.

### Changes

- Set a maximum height on the filter section before scrolling
- Let individual filters expand as much as needed

### Implementation Notes

- The solution here with `calc` is simplistic. If the header gets taller in the future then we may need to brush it up.

### Test Plan

1. Log in as `statesubmitteractive`. Create at least 8 submissions.
2. Go to the package dashboard and click the Filter button. Ensure that the "Reset" button is at the bottom of the viewport regardless of whether the accordions are expanded, and ensure that the accordions do not scroll internally.

Screen Shot 2021-12-08 at 2 14 56 PM

### DIFF
--- a/services/ui-src/src/index.scss
+++ b/services/ui-src/src/index.scss
@@ -2218,8 +2218,12 @@ aside#faq-contact-info-box {
     }
 
     .filter-accordion {
-      flex: 1;
+      flex: 0 1 calc(100vh - 325px);
       overflow-y: auto;
+
+      .ds-c-accordion__content {
+        max-height: unset;
+      }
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
Style changes look appropriate... endpoint is: https://d297dpjl21edyv.cloudfront.net

I added enough packages to see that the filter pane does not force a scroll to use.

Approved!

PR #563, opened 2021-12-20
State should not be able to withdraw a package in OneMAC in Specific statuses

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-11950
Endpoint: https://d2k3etjprsxw0q.cloudfront.net/

### Details
Ad we add gate-keeping to the package actions, we need a structured way to enforce that the proper actions are assigned to the proper states.  While it was tempting to use xstate for this, research showed that this approach would lead to every line of the table creating a state machine.  So instead, we used the CONFIG structures used for the forms to differentiate between packages.

### Changes
- added the concept of Actions to the package configurations
- created a default status/actions config, because most packages work the same
- standardized the PackageList menu item creation

### Implementation Notes
Used the existing config structure to carry the package-specific triage.  Currently, all packages are the same, so they use a default status/action mapping.

### Test Plan
1. Login as statesubmitter@nightwatch.test to see list of packages (is pre-set with the statuses)
2. Verify that the Withdraw menu is available according to the following:

Status. (can withdraw):
- Unsubmitted. (NO)
- Submitted. (YES)
- Package In Review. (YES)
- RAI Issued. (YES)
- Package Approved. (NO)
- Package Disapproved. (NO)
- Package Withdrawn. (NO)
- Waiver Terminated. (NO)

### DIFF
--- a/services/app-api/one-seed.json
+++ b/services/app-api/one-seed.json
@@ -1,0 +1,186 @@
+[{
+    "pk": "MD-13-1122",
+    "sk": "spa",
+    "submitterId": "us-east-1:afa582ca-4e4c-4d3b-be9b-d2dbc24c3d1a",
+    "changeHistory": [
+     {
+      "componentType": "sparai",
+      "componentTimestamp": 1639696189171,
+      "submitterName": "StateSubmitter Nightwatch",
+      "componentId": "MD-13-1122"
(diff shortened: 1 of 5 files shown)


g-s-k (approved):
nice

PR #578, opened 2022-01-10
OY2-14657: limit data return on Dashboard

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-14657
Endpoint: https://drf6yzg7zrgjt.cloudfront.net/

### Details

Reduce the size of the returned list from `getAllByAuthorizedTerritories`.

### Changes

- Limit keys returned on each change request object when querying Dynamo

### Implementation Notes

- May need to update this projection if we need more keys in this view

### Test Plan

1. Log in as `statesubmitteractive`. Smoke test the "old" dashboard view to ensure all links still work and RAI responses still process correctly.
2. Log in as `helpdeskactive` and smoke test again.

### DIFF
--- a/services/app-api/getAllByAuthorizedTerritories.js
+++ b/services/app-api/getAllByAuthorizedTerritories.js
@@ -5,6 +5,13 @@ import { RESPONSE_CODE } from "cmscommonlib";
 import getUser from "./utils/getUser";
 import { getAuthorizedStateList } from "./user/user-util";
 
+const commonQueryConfig = {
+  TableName: process.env.tableName,
+  ExpressionAttributeNames: { "#type": "type", "#user": "user" },
+  ProjectionExpression:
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
I see the code climate for the actions says this is a 100% covered change, but this script is only 6.58% covered according to the Unit Test Coverage Report.  It would be nice to see more unit testing added here.

Doing dumb-dumb comparison between the other PR and this one, helpdesk account takes 1.24 s and sends 2.3 MB for the Search Bar label story and 615ms and transfers 367KB for this one, so I'd say that's an improvement (barring any unknown browser or network caching and I didn't verify the data sets are the same, like I said, dumb-dumb)

Clicked around as statesubmitteractive for MI and VA and statesubmitter@nightwatch for MD and helpdesk for all and did not see anything amiss.

Nice use of Projections!  Approved

PR #579, opened 2022-01-11
put phone number unit test and refactor

jjames521:
Story: https://qmacbis.atlassian.net/browse/OY2-14301

### Details

Unit test for putPhoneNumber

### Changes

- Try catch block in putPhoneNumber to refactor
- Unit test add

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2

### DIFF
--- a/services/app-api/putPhoneNumber.js
+++ b/services/app-api/putPhoneNumber.js
@@ -29,9 +29,14 @@ const validateInput = (input) => {
  */
 export const main = handler(async (event) => {
   let input = event.body;
-  if (!isObject(input)) input = JSON.parse(input);
+  try {
+    if (!isObject(input)) input = JSON.parse(input);
 
-  if (!validateInput(input)) {
+    if (!validateInput(input)) {
+      throw "Validate input returned undefined";
+    }
+  } catch (e) {
+    console.log("Error is", e);
     return RESPONSE_CODE.VALIDATION_ERROR;
   }
 
--- a/services/app-api/putPhoneNumber.test.js
+++ b/services/app-api/putPhoneNumber.test.js
@@ -11,10 +11,10 @@ dynamoDb.update.mockImplementation(() => {
 
 const mockObject = {
   statusCode: 200,
-  body: JSON.stringify({
+  body: {
     id: "",
     phoneNumber: "",
-  }),
+  },
   headers: {
     "Access-Control-Allow-Origin": "*",
     "Access-Control-Allow-Credentials": true,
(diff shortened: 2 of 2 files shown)


services/app-api/putPhoneNumber.test.js
```diff
@@ -1,8 +1,103 @@
...
+    id: "statesubmitteractive@cms.hhs.local",
+    statusCode: 200,
+    body: RESPONSE_CODE.VALIDATION_ERROR,
+    headers: {
+      "Access-Control-Allow-Origin": "*",
+      "Access-Control-Allow-Credentials": true,
+    },
+  };
+
+  await expect(main(mockObject))
```
kristin-at-theta, line 32: await should no longer be needed, please remove :)

services/app-api/putPhoneNumber.test.js
```diff
@@ -1,8 +1,103 @@
...
+    .resolves.toStrictEqual(expectedResponse)
+    .catch((e) => {
+      RESPONSE_CODE.VALIDATION_ERROR, e;
```
kristin-at-theta, line 35: I think we want to console.log all testing errors, so please make this catch block the same as the final one: 
    .catch((e) => {
      console.log("error", e);
    });

kristin-at-theta (changes requested):
I know this was a lot of work and it is nicely done.

I do think the final implementation could benefit from some cleanup :) 

changes requested: 

- mockObject and expectedResponse are duplicated in every test, can make them global and edit the differences found in each test.
- the await expect should no longer be needed now that you are mocking the dynamodb function
- I like that you catch the test errors, but I feel like they need to be logged as well.

These are minor updates that I believe will enhance the value of the code, but if we are running into timeline issues (or you just want to be done with it), none of my requests are showstoppers and ask and I will approve :)

services/app-api/putPhoneNumber.js
```diff
@@ -29,9 +29,14 @@ const validateInput = (input) => {
  */
 export const main = handler(async (event) => {
   let input = event.body;
-  if (!isObject(input)) input = JSON.parse(input);
+  try {
+    if (!isObject(input)) input = JSON.parse(input);
```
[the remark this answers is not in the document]
jjames521: Thank you!

services/app-api/putPhoneNumber.test.js
jjames521: Done!

g-s-k (approved):
looks good!

PR #580, opened 2022-01-13
OY2 14464 -- Package Dashboard: Separate Tab for Waivers and SPAs

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-14464
Endpoint: https://d1bdpxawgpffyj.cloudfront.net/

### Details
Separating the dashboard into SPA-specific items and Waiver-specific items

### Changes

- Added the "tabs" to the PackageList
- Modified getMyPackages to filter on type
- switching tabs changes background data

### Implementation Notes

- Not optimizing at this time, using the buttons causes a reset of the data.  Should probably look into using cached data, but not sure how easy that is to do in dynamoDB

### Test Plan

1. Log in as a User that sees the Package List
2. Go to Packages, should default to only SPAs showing
3. Verify columns are the ones specific to SPAs
4. Verify filters work on the data shown
5. Verify only SPAs are shown (and the right list based on territories and user role)
6. Click Waiver tab
7. Verify columns are the ones specific to Waivers
8. Verify filters work on the data shown
9. Verify only Waivers are shown (and the right list based on territories and user role)


g-s-k (approved):
looks good!

PR #582, opened 2022-01-13
OY2-13094: add filter for territories

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-13094
Endpoint: https://d1knihpd5y840j.cloudfront.net/

### Details

Add a filter in the Filter pane for the State column.

### Changes

- Make State column filterable
- Implement filter UI for territories

### Implementation Notes

- N/A

### Test Plan

1. Log in as `helpdeskactive` and navigate to Packages.
2. Click on the Filter button and verify that the State column now has a corresponding entry in the filter pane.
3. Open the filter for State and ensure that it works in the following scenarios:
    - Single state selected that **is not** present in the list
    - Single state selected that **is** present in the list
    - Multiple states selected that **are not** present in the list
    - Multiple states selected that **are** present in the list
    - Multiple states selected where **some are** present in the list
    - Clearing the input with the X button
    - Clearing the input with the Reset button

### DIFF
--- a/services/common/index.d.ts
+++ b/services/common/index.d.ts
@@ -34,7 +34,9 @@ export const latestAccessStatus: (
   territory: string
 ) => USER_STATUS;
 
-type SelectOption = { label: string; value: string };
+export type SelectOption = { label: string; value: string };
+
+export const territoryList: SelectOption[];
 
 export namespace ChangeRequest {
   type TransmittalNumberInfo = {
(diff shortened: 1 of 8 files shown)


kristin-at-theta (changes requested):
alignment of Search bar is off on my screen:
Screen Shot 2022-01-14 at 12 35 10 PM

Sorry if the things I requested during pairing had that side effect!!!

Everything else looks good and works as expected :)

kristin-at-theta (approved):
back in alignment !

Approved!

PR #591, opened 2022-01-19
OY2-14941: increase test coverage from bottom up

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-14941
Endpoint: TBD

### Details

Focusing on files with no coverage rather than largest files this time.

### Changes

N/A

### Implementation Notes

- Also modified test that was failing locally to stop doing that
- Removed a file that seems to be unused

### Test Plan

1. See bigger number


kristin-at-theta (approved):
Nice!


Approved!

PR #597, opened 2022-01-21
Regression Automation Maintenance: Increased viewport

Valencia2019:
Changes: Increased viewport.
Reasoning: This is the fix for the actions button sometimes not being visible in viewport. That is happening because the computer doesn't know to scroll right to see the object. The simplest solution that makes sense is increasing the viewport. It is now set to that of a Macbook-13 inch.


kristin-at-theta (approved):
Hey!  I have a MacBook like that!

Approved!

PR #610, opened 2022-01-27
run seed as its own step on every deploy

g-s-k:
Story: N/A
Endpoint: TBD

### Details

This will force any data present in our seed files to match what they say on deploy, even if it is not the first deploy on a branch. Any data added ad hoc after deploying will be preserved, and this step will not run in VAL or PROD.

### Changes

- N/A

### Implementation Notes

- N/A

### Test Plan

1. N/A


kristin-at-theta (approved):
Yup, looks good!

Approve!

PR #626, opened 2022-02-02
OY2-12679 and OY2-14752: role change requests

g-s-k:
Stories:
- https://qmacbis.atlassian.net/browse/OY2-12679
- https://qmacbis.atlassian.net/browse/OY2-14752
Endpoint: https://dt3um01ognfr.cloudfront.net/

### Details

Enable users to request a different role.

### Changes

- Add "request a role change" link in User Profile dropdown that takes you to the signup page
- Allows you to request a role different from your current role
- If approved, revokes access to your current role

### Implementation Notes

- Users now live in the `onemac-*-one` table under a versioned schema.

### Test Plan

TBD


kristin-at-theta (approved):
Approved!

PR #628, opened 2022-02-04
Fix av

Dark-Knight-000:

### Details

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2


kristin-at-theta (approved):
I guess it looks ok?? I don't see anything glaringly wrong, but I have very little familiarity with the AntiVirus definitions and will have to defer to Platform team :)   

What I do see is that this should not adversely affect the parts of the application I *do* know about, so .... Approved!

PR #630, opened 2022-02-08
Table & Other Factoring

jjames521:
Story: N/A
Endpoint: https://d38tlyg2ez4nra.cloudfront.net/

### Details

This is meant to be a clean up of previous old icons etc. and refactor of the way we handle resize and overflow of columns on table.  Along with this a utilization of React-Spring for animations on mount and unmount.

### Changes

Changes:
- Margin padding for focus being visible on top and bottom of filter accordions
- padding added for dropdown when list is smaller, preventing overflow of date range picker
- My Account animation add - icon flips up and menu appears with keyframe fade in and fade out
- Introduction of React-Spring for mount and unmount useTransition hooks, filter and hamburger utilize this, subtle fade in and fade out
- Add in of show/hidden columns caret flip up
- Default columns were changed Waiver Family # and Expiration Date defaultHidden
- ID sticky table, caret sticky, child components when expanded sticky ID added,  also added in styles for background to be consistent
- New Submission icon change over, Export CSV icon removed as was not indicative of the same actions as New Submission, avoiding confusion

### Implementation Notes

- Unit tests updated
Globals called to skip timing
```
import { Globals } from "@react-spring/web"

Globals.assign({
  skipAnimation: true,
})
```

-to skip any animation delay, waitFor utilized as well.

### Test Plan

1. Log in with _statesubmitteractive_, click on filter, should see a subtle fade in animation, try to utilize keyboard tabs to see focus states.  
2. Close filter and check out My Account for a fade in and caret animation.
3. Click Waivers and Check out Show/Hide Columns, should be able to see that Waiver Family # and Expiration Date are defaulted as hidden.  See if caret is animated to be a twisty.
4. Resize browser window to 930 or below, check out Hamburger Menu and see subtle fade in animation pull the window out past 940 and see it disappear.
5. Check New Submission button, should have new + icon instead of circle +
6. Resize window and scroll horizontally on table can also enable columns from hide/show columns to increase this length.
7. Should see that ID and if applicable the caret is sticky
8. Log out of _statesubmitteractive_ and into _systemadmintest_ and check out waiver tab, should see an applicable active caret to test, expand and see that the sticky columns have a light blue background and are sticky when horizontally moving across the table.
9. Check export to excel on all main nav components doesn't have an icon any more.


kristin-at-theta (commented):
in the filter pane, tab works to enter and exit the date picker downwards, but shift-tab will go into the date picker, yet not leave it.

Hitting "return" key on the expanders correctly expands/contracts filter menus.

Probably out of scope for this PR, though important for 508 compliance...
Hitting "return" key with a checkbox selected does not check/uncheck.  Hitting "return" key with date picker enabled does not open the picker - can type dates into picker, but cannot set the filter using keyboard.  
State picker works as expected using keyboard.

The animations I can see look good - I can't see the hamburger menu one or the filter pane, but that's not super-surprising

horizontal scrolling on table and default columns look good, colors are maintained, but.... I am able to scroll far enough right to remove the caret and then cannot expand/close that line item until I scroll back left. 

All in all, looks great and I defer to your judgement on if any of the things I found warrant changes or can be fixed later (I have heard we have an entire epic devoted to 508, so those can probably be tackled then...) 

However, you are still down -0.1 % in the test coverage - which would be fine *except* that I really don't to lose our 80% before the week is out :)

kristin-at-theta (approved):
Yeah!  see the general comment for things I observed, but from what I see, this can be...

Approved!

PR #640, opened 2022-02-15
OY2-15620: Updated SPA Package Details View - Card View with Actions

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-15620
Endpoint: https://d2ejxjm0qtmoph.cloudfront.net

### Details
The first story to clean up our package details pages!  The story is specific to SPA details, but the page itself is component-neutral-yet-aware (right now)

### Changes
- Add a two-column card to top of details
- include allowed package actions in card
- no longer display extra details

### Test Plan
1. Log in as a user that can see SPA package details
2. View a SPA and verify the card matches what is asked for in the story

### DIFF
--- a/services/ui-src/src/DynamicRoutes.tsx
+++ b/services/ui-src/src/DynamicRoutes.tsx
@@ -163,13 +163,13 @@ export default function DynamicRoutes() {
         ))}
       {userRoleObj.canAccessDashboard && (
         <>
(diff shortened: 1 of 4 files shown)


g-s-k (approved):
i am not super appalled by the level of copy/paste here but it's not my favorite. good enough for now. also, i know that the AC's don't say what to do when there are no actions available...but a heading with nothing below it looks kind of janky. 🤷

PR #659, opened 2022-02-25
allow dev login in VAL

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-16219
Endpoint: TBD

### Details

Allowing test users to access OneMAC in the VAL environment.

### Changes

- Set the flag in `master` branch builds in addition to `develop` and feature branches
- Users will now see the Packages link and CMS Reviewer role if they log in as a test user in VAL.

### Implementation Notes

- All other features we choose to flag with the `ALLOW_DEV_LOGIN` variable will be visible in VAL going forward.

### Test Plan

N/A

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -121,8 +121,8 @@ jobs:
       - name: set path
         run: |
           echo "PATH=$(pwd)/node_modules/.bin/:$PATH" >> $GITHUB_ENV
-      - name: Set dev login flag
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
yup, that's what I thought would be the thing - though I hope you also did some poking around to verify that the flag is used in the right spots and there are no other blocks on the stage="master" since stage name is something known ....

Can't really test this until it gets to VAL though.... so - Approved!

PR #662, opened 2022-02-28
created new actions by status array for RAIs

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-16471
Endpoint: https://d1shu6z17bkree.cloudfront.net
### Details

As a State Submitter/Admin, I should not have the ability to withdraw RAI Responses in OneMAC so that all RAI response withdrawals can be handled outside of OneMAC.

### Changes

- Created new action mapping for rai statuses
- Assigned new mapping to RAI type packages

### Implementation Notes

- NA

### Test Plan

1. Login as State Submitter
2. Create a waiver
3. Create an RAI response on the waiver
4. Navigate to package view and ensure there is no "Withdraw Package" option on RAI Response


g-s-k (approved):
nice work!

PR #669, opened 2022-03-01
remove attachments from the top level updates

kristin-at-theta:
Story: More of a Figma observation
Endpoint: https://d2w5ecfqlh1f36.cloudfront.net/

### Details
When we created the data model, the Figmas depicted all package attachments rolled up into one attachments section, so that's how they were stored in the database.
New Figmas and stories are clearly different in that regard, with a "Base Supporting Documentation" section and Attachment sections under other component detail areas.

This PR removes the rollup of the attachments.

### Changes
- removed attachments as a top level attribute to be updated when a new submission is received

### Test Plan
1. Log in as submitter user
2. Submit a child component to a parent component viewable in OneMac
3. verify that the attachments are not in the "Base Supporting Documentation" section


bflynn-collabralink (approved):
Seems like this goes hand in hand with https://qmacbis.atlassian.net/browse/OY2-16322

PR #670, opened 2022-03-02
implement vertical nav for detail view

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-14656
Endpoint: TBD

### Details

Implement a vertical navigation element in our package/component detail view.

### Changes

- Vertical tabbed navigation on the left side of the detail view

### Implementation Notes

- Some typings cleanup in both DetailView and its dependencies, and PortalTable

### Test Plan

1. Log in as `statesubmitteractive` and go to Packages
2. Click on a package or child component
3. Verify that the nav is present and looks sort of like the Figma. Note that there should only be one item, so functionality will not really be testable until the next story comes through.


kristin-at-theta (approved):
Focus seems weird with the purple box on the expanding items, but no ability to select the lower choices without the mouse (I tried arrow keys, tab, and enter).  Not sure what it should be, though, so hopefully there will be guidance from the 508 tools.

Looks like the Figma (though sparse because of missing pages), it looks like the code is in place to manage switching between pages, so I suppose we can keep that AC about switching pages even though it is not-demonstrable?

test coverage could be better, but meets requirements

Approved

PR #681, opened 2022-03-04
OY2 16707: CMS Users Denied a CRA Role loses Read Only Access to OneMAC

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-16707
Endpoint: https://d36mye51rm55p.cloudfront.net/

### Details
The CMS default user should be added as a new role in OneMAC in order for approval/revoking/default/fallthrough behaviors to match the other role types.

Also, a lot of problems arose because the default users were being shown an empty submissions list on their first access, which prompted them to ask for the CRA role (and then get denied)

### Changes
- added defaultcmsuser as a role
- Created label for role as "CMS Read-only User"
- "Approver" is CMS System Admin
- added a second query after a pause for the roleList if the first query comes up empty
- reload the app if the user was created.

### Implementation Notes
- had to do some extra checking to get rid of the "no submissions" screen
- so.... fixing the test coverage made it easy to find/fix the csv button, though I haven't tested that yet.....

### Test Plan

1. Login as an unregistered user
2. View the default CMS features (Dashboard, user profile)
3. Request the CMS Role Approver Role
4. Continue to see default CMS features
5. Use system admin to grant CRA role
6. See CRA features
7. Use system admin to revoke CRA role
8. See default features again

### DIFF
--- a/loadTestUsers.py
+++ b/loadTestUsers.py
@@ -125,7 +125,7 @@ def seed_cognito(test_users, user_pool_id, password):
         elif user["role"].startswith("state"):
             role = "onemac-state-user"
         elif user["role"].startswith("cms") or user["role"] == "systemadmin":
-            role = "onemac-cms-user"
+            role = "none"
         elif user["role"] == "helpdesk":
             role = "onemac-helpdesk"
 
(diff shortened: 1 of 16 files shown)


g-s-k (approved):

bflynn-collabralink (approved):
Looks good assuming those two user tests will get cleaned up.

PR #694, opened 2022-03-14
oy2-15452 - set 90 day as "Pending" on package/detail page for in rev…

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-15452
Endpoint: https://d34jicl8rg41gf.cloudfront.net/

### Details

set 90 day property as "Pending" on package/detail page for in review items

### Changes

- moved logic for determining textual 90 day clock from  PackageList.js to changeRequest.js so it is reusable
- use new method for 90 day value on PackageList and DetailView

### Implementation Notes

- I left the common method in changeRequest unformatted when it is a date value because each usage may want to determine its own format; furthermore the current format on packagelist and detailview are slightly different

### Test Plan

1. Login as state submitter
2. Navigate to Package List
3. Verify any package in status "In Review" has the 90th day column has value "Pending"
4. Navigate to detail view from package list for a package in "In Review" status
5. Verify 90th day value in detail card has value "Pending"

### DIFF
--- a/services/common/changeRequest.js
+++ b/services/common/changeRequest.js
@@ -117,6 +117,24 @@ export const getWaiverRAIParent = (inId) => {
   return TYPE.WAIVER_BASE;
 };
 
+export const get90thDayText = (currentStatus, clockEndTimestamp) => {
+  switch (currentStatus) {
+    case ONEMAC_STATUS.RAI_ISSUED:
+      return "Clock Stopped";
+    case ONEMAC_STATUS.APPROVED:
+    case ONEMAC_STATUS.DISAPPROVED:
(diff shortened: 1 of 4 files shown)


kristin-at-theta (approved):
So, this is implemented per the AC, but now there is no status for which the actual 90th day shows?  Before, the requirement was that the status was "Pending" ONLY if the timestamp wasn't available for packages in "Submitted" or "Package In Review".  I suppose this is a change because of not getting SEATool yet.

Approved!

bflynn-collabralink:
> 

I noticed this too and was something I wanted to raise during demo. Kristin with the keen eye!

PR #706, opened 2022-03-17
OY2-16976: Base Waiver: Verbiage and Validation for Submission

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-16976
Endpoint: https://d2ajx5kog67s2l.cloudfront.net/

### Details
Updating forms for package view means they should be accessed from and return to the package dashboard.

### Changes
- returned the legacy New Submission functionality to previous incarnation
- created new triage choice pages for the new form versions

### Test Plan
1. Login as a state submitter
3. Click "New Submission" button
4. Click "Waiver" then "Base Waiver"
5. Submit a Base Waiver using the form
6. after clicking submit, notice that you are returned to the package dashboard and are shown the "Submission success" message


bflynn-collabralink (approved):
Looks great! You make it look so easy but i know better. :)

One comment would be should we consider loading the Legacy SPA forms into the Triage group for the forms we havent yet replaced? Kind of consider the current version is what we have until we replace it? Maybe, maybe not.

Also noticed after submission of a waiver and returned to the waiver tab the route works well but the url is now fixed on /packagelist-waiver even when changing tabs, just dont know if that will cause any issues in future other than the visual disconnect.

kristin-at-theta:
I deliberated over adding the legacy forms, and ended up deciding it was better to separate them fully at least at first.  Partly because ryan is new and I don't want to force him to figure out which things have been converted and which haven't.  Plus.... in this next story, I plan to really rip apart and restructure the forms if I can, so better to only have ONE new form to retrofit :)  

Good find on the url inconsistency... I didn't re-run the tabs through the router and we probably should.  That would follow the same pattern we are establishing for those left-nav tabs, too.  I'm going to punt that for later though, since I'm not exactly sure how to go about it and there's that 5 pointer staring at me :)

Dark-Knight-000:
I agree with the decision on the separating the form instead of adding the
legacy.

PR #721, opened 2022-03-24
OY2-16794: add CC emails for CHIP submissions

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-16794
Endpoint: https://dr2di96l15e3f.cloudfront.net/

### Details

Pick up GH secret in prod to CC CHIP emails to CHIP Project Officers.

### Changes

- Add CC capability to email lib
- Add CC addresses from env var on CHIP-related actions

### Implementation Notes

- Env var passthrough is very clunky as is. Maybe we should try to simplify that in the future.

### Test Plan

1. Go into the AWS console and edit the environment variable `chipCcEmail` for the function `...-submit` to an email address you can access. If you are going above and beyond and testing named email addresses, make sure you **double quote** the name portion if it contains a comma.
2. Log in as `statesubmitteractive` and submit a CHIP SPA or CHIP SPA RAI. Check that email and verify that it gets CC'd on the receipt email.
3. Repeat step 1 for `...-withdraw-package`.
4. Withdraw a CHIP SPA from the package dashboard. Verify that you get another CC'd email.


kristin-at-theta (changes requested):
So the semicolon delimiter threw me for a loop until I looked at the story... and that email address list will not work as it is written - the commas are not contained within " characters and email standard is comma-delimited individual addresses, so those names will get broken up and attempted as separate email addresses.

Its been a while since I was an "email guru" so I went ahead and tested using:
`Phelps, Sandra (CMS/CMCS) <k.grue@theta-llc.com>; Singleton, Shakia (HRSA) <kgrue@collabralink.com>; `
... submit fails with a pass-through error of {"error":"Illegal address"}.

When I quoted the names, it got farther, but I think it tried to send to the space after the last ; ?  I got:
{"error":"Invalid email address  ."}

g-s-k:
> So the semicolon delimiter threw me for a loop until I looked at the story... and that email address list will not work as it is written - the commas are not contained within " characters and email standard is comma-delimited individual addresses, so those names will get broken up and attempted as separate email addresses.

oof. i only tested with plain addresses, that'll show me

kristin-at-theta:
I will approve as it fulfills on the AC, but I am not fond of the fact that a failure in the cc emails causes the submission to happen, but an error is received on the form screen and it looks like you need to resubmit the form.  This feels like a disaster waiting to happen and I encourage us to be very careful when deploying to PROD as that will be the first time the list of cc emails as it exists in PROD will be tested.....

kristin-at-theta (approved):

PR #722, opened 2022-03-25
OY2-17314: Correct "Last Modified Date" for CMS users

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-17314
Endpoint: https://dlshp6xcc049a.cloudfront.net/

### Details
Jaime noticed that some users have a last modified date in the far far future.  Looks like the initial creation of the read-only CMS user dynamodb item used milliseconds (Javascript) instead of seconds (Unix).

Turns out, the entire application uses milliseconds and ONLY last modified in the user table was in seconds.  Converting that timestamp to use milliseconds to match the rest of the database AND the front end (which was multiplying by 1000 every query... yikes!)

### Changes
- added a check to see if the date was sent incorrectly and fixes the timestamp into the proper units.
- migrate updated to convert existing user item dates with the incorrect timestamp units

### Test Plan
1. Use AWS console to view items in "onemac-develop-one"
2. Query your github email address, if items, then delete those items (all of them) 
3. Navigate to develop endpoint: https://d2dr7dgo9g0124.cloudfront.net/
4. Use your github email address (and standard pw) to log into OneMAC
5. Logout as github user
6. Login as system admin and see the crazy future date
7. Switch to feature branch database: "onemac-oy2-17314-one"
8. Query your github email address, if items, then delete those items (all of them) 
9. Navigate to: https://dlshp6xcc049a.cloudfront.net/
10. Use your github email address (and standard pw) to log into OneMAC
11. Login as system admin and see today's date


bflynn-collabralink (approved):
LGTM

PR #730, opened 2022-03-31
OY2-17369 && OY2-11205: Fix multi select component

g-s-k:
Stories:
- https://qmacbis.atlassian.net/browse/OY2-17369
- https://qmacbis.atlassian.net/browse/OY2-11205

Endpoint: https://d6xzribkkklsx.cloudfront.net/

### Details

Don't keep using a janky library for our state selection component.

### Changes

- Replace `react-dropdown-select` with `react-select` (which we already use elsewhere)

### Implementation Notes

- `react-dropdown-select` has been removed from our dependency list

### Test Plan

1. Log in as `statesubmitterunregistered` and select "State Submitter". Try to reproduce the bug described in OY2-11205.
2. Click Cancel or the back button, select "State System Admin", and try again.
3. Log out, then log in as `statesubmitteractive` and go to your profile. Click "Add State" and try to reproduce it there.


kristin-at-theta (changes requested):
I really like that it is an obvious single-select for State Admin Users!

Though... when I tried to submit, I got:
Could not create new user: TypeError: e.map is not a function

kristin-at-theta (approved):
Looks like it works now.  I definitely like this component better than the other one.

A couple things:
- seems weird that it keeps the expander icon thingie, it doesn't appear to do anything when you click it, but odd that it is there at all?
- the cancel button turns into a gray disabled button when the Submit button is clicked, I believe this was because we disable them to prevent multiple clicks, but I think the style is perhaps off?

Anyway, neither of those is serious (to me), so with that - Approved!

jjames521:
> Looks like it works now. I definitely like this component better than the other one.
> 
> A couple things:
> 
> * seems weird that it keeps the expander icon thingie, it doesn't appear to do anything when you click it, but odd that it is there at all?
> * the cancel button turns into a gray disabled button when the Submit button is clicked, I believe this was because we disable them to prevent multiple clicks, but I think the style is perhaps off?
> 
> Anyway, neither of those is serious (to me), so with that - Approved!

One thing I'm seeing is ability to tab forward to actually submit, might hang us up on any 508 compliance.

jjames521 (dismissed):
One thing I'm seeing is ability to tab forward to actually submit. It works for situations where you can multi-select like in the profile but not in Request a New Role. That might hang us up on any 508 compliance.

PR #735, opened 2022-04-06
OY2 13089: AWS Security Hub GitHub Collector - OneMAC

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-13089
Endpoint: https://d3u89wtezmfj6t.cloudfront.net/

Details
Follow the instructions in https://github.com/CMSgov/macpro-quickstart-serverless/pull/319 to add the AWS Security Hub Findings to the GitHub Issues and eventually, a slack channel

Changes
- added a new deploy-support workflow
- added the .sechub lambda

### Test Plan

1. Nothing should change in the application
2. can look at the lambda running, but it won't get any findings until merged with develop....


.github/package.json
```diff
@@ -15,12 +15,12 @@
...
   },
   "author": "",
   "license": "CC0-1.0",
   "bugs": {
-    "url": "https://github.com/CMSgov/macpro-quickstart-serverless/issues"
+    "url": "https://github.com/CMSgov/onemac/issues"
   },
-  "homepage": "https://github.com/CMSgov/macpro-quickstart-serverless#readme"
+  "homepage": "https://github.com/CMSgov/onemac#readme"
 }
```
g-s-k, line 26: i thought we wanted to delete this file...

g-s-k (approved):
is there a reason why we are using a hidden directory for the lambda? just curious

kristin-at-theta:
We are using the hidden directory because that is how the QuickStart does it and I'm reluctant to stray too far from their pattern....

Yeah, that branchNameValidation.sh. really feels like someone needed to placate code climate, though I didn't know it looked at those sorts of files...

As to deleting .github/package.json - this is where I am not aware enough of how the packages work... I don't think .github uses node or packages, but somehow dependabot finds them? and well.... if you have seen my attempts to mess with the packages, you will see a LOT of failed "cleanups" that I've given up on....

Dark-Knight-000 (approved):

PR #748, opened 2022-04-15
OY2-17605: implement "skip to content" link

g-s-k:
Story: https://qmacbis.atlassian.net/browse/OY2-17605
Endpoint: TBD

### Details

Add a "skip to content" link for 508 compliance and general keyboard navigation.

### Changes

- "Skip to content" should be the first available tab destination when using a screen reader, and activating it should take your focus past the nav bar

### Implementation Notes

- Also fixed an existing styling issue causing the homepage to scroll horizontally

### Test Plan

1. Open the home page. Hit `Tab` and ensure that the "Skip to content" link appears on screen. Hit `Enter` and verify that you are scrolled past the nav bar, and that a subsequent `Tab` key will not focus the links there.
2. Repeat with Google's "Screen Reader" extension enabled. Ensure that the text of the link is read aloud prior to anything else on the page.


kristin-at-theta (approved):
Was waiting to hear the Jesse changes were in - guess they are :) 

Approved!

PR #750, opened 2022-04-19
Oy2 17593 fix

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-17593
Endpoint: https://ducrroevj7t96.cloudfront.net

### Details

Add a bucket and turn on logging for cloudfront

### Changes

- updated ui/serverless.yml to add bucket/bucket policy and configure cloudfront distribution to output access logs to the bucket

### Test Plan

Access site and ensure logs are created in bucket with name ui-{stage}-cloudfront-logs-{accountId}
Ensure bucket has at rest encryption enabled and is only accessible by https

### DIFF
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -76,8 +76,8 @@ resources:
               Principal: "*"
               Action: "s3:*"
               Resource:
-                !Sub ${S3Bucket.Arn}/*
-                !Sub ${S3Bucket.Arn}
+                - !Sub ${S3Bucket.Arn}/*
+                - !Sub ${S3Bucket.Arn}
               Condition:
                 Bool: 
                   "aws:SecureTransport": "false"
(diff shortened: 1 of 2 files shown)


kristin-at-theta (changes requested):
Looks good, although you'll need that second "Deny" (or figure out how to do the two resources for the one "Deny" on the ssl layer - I think the /* only does the stuff in the bucket and you need to do a / as well to get the main level of the bucket...

bflynn-collabralink:
> Looks good, although you'll need that second "Deny" (or figure out how to do the two resources for the one "Deny" on the ssl layer - I think the /* only does the stuff in the bucket and you need to do a / as well to get the main level of the bucket...

good catch! updated with second deny policy for the root

kristin-at-theta (approved):

PR #755, opened 2022-04-21
OY2-17614 && OY2-17653: ARIA labels for various buttons

g-s-k:
Stories: 
- https://qmacbis.atlassian.net/browse/OY2-17614
- https://qmacbis.atlassian.net/browse/OY2-17653
Endpoint: https://d1031h39uabus0.cloudfront.net/

### Details

Provide helpful descriptions for three types of button in the OneMAC application.

### Changes

- Back buttons are now `aria-label`ed with "Back to previous page"
- Actions buttons on Submission Dashboard are now `aria-label`ed with "Actions for <transmittal number>" as on Package Dashboard
- File upload buttons are now `aria-label`ed with "Add File(s) of type <upload type>"
- File upload buttons now receive the same focus styles as all other buttons and form fields

### Implementation Notes

- NB: Use `@include focus-styles;` if you need the stock CMSDS focus styles on anything. That is what they use internally.

### Test Plan

1. Ensure you have the Google Screen Reader extension installed. It does not always need to be enabled, but you will need to use it for each test case.
2. Log in as `statesubmitteractive` and go to the legacy Dashboard. Enable the screen reader and tab onto one of the Actions buttons. Ensure that it reads out "Actions for <transmittal number>" when it is focused.
3. Click the "New Submission" button. Tab onto the back button and ensure that you hear "Back to previous page".
4. Select one of the submission forms. Tab downward to the file upload section. When you reach the Add File buttons, ensure that they get the standard purple border, and that rather than just hearing "Add File" you hear "Add files of type <upload type>".

### DIFF
--- a/services/ui-src/src/components/FileUploader.js
+++ b/services/ui-src/src/components/FileUploader.js
@@ -255,6 +255,9 @@ export default class FileUploader extends Component {
           </td>
           <td className="uploader-input-cell">
             <label
+              aria-label={`Add file${
+                uploader.allowMultiple ? "s" : ""
+              } of type ${uploader.title}`}
               className={
                 isDisabled
                   ? "uploader-input-label-disabled"
--- a/services/ui-src/src/components/PageTitleBar.js
+++ b/services/ui-src/src/components/PageTitleBar.js
@@ -54,6 +54,7 @@ const PageTitleBar = ({
         <div className="title-bar-left-content">
           {enableBackNav && (
             <Button
+              aria-label="Back to previous page"
               id="back-button"
               data-testid="back-button"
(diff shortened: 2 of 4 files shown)


kristin-at-theta (approved):
Testing this was a bit surreal, and there are plenty of gotchas, especially around the popup menu items and how they don't read well and also don't seem to be actionable via the keyboard.... but that is not what we care about here!

It correctly informed me about the "Back" buttons and also identified the Actions for #. when it actually was aware of the table row... I did find it wasn't hard to get into selecting the table without it knowing what row to talk about?  I guess we'll learn more about that as we go.

Anyway, nice clean work - thanks !
Approved

PR #756, opened 2022-04-25
Oy2 17566tests

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-17566

### Test Plan
Re-enabled this test:

```
Feature: State should not be able to withdraw RAI Responses in OneMAC
    Scenario: Can not withdraw Waiver RAI Response
        Given I am on Login Page
        When Clicking on Development Login
        When Login with state submitter user
        Then click on New Submission
        And Click on Waiver Action
        And Click on Waiver Action under Waiver Type
        And Click on New Waiver under Action type
        And Click on 1915 b 4 FFS Selective Contracting waivers under Waiver Authority
        And Type Unique Valid Waiver Number With 5 Characters
        And Upload 1915 b 4 file
        And Click on Submit Button
        And verify submission Successful message
        And click on Waiver Respond to RAI
        And Add file for Waiver RAI Response
        And Click on Submit Button
        And verify submission Successful message after RAI
        And click on Packages
        And click on the Waivers tab
        And search for Unique Valid Waiver Number with 5 Characters
        And wait for parent row expander to be enabled
        And click parent row expander
        And verify actions button on the child row is disabled

       # Scenario: Can not withdraw SPA RAI Response
       #  this is not possible in package dashboard since RAI responses
       #  don't appear on the SPA tab.
```


Valencia2019:
Checked the two tests failing. They're happening because of data differences in the branch environment. I ran the tests against develop and they are both passing so it should be fine.

kristin-at-theta (approved):
Tests look good, changes contained within testing areas.

Deployment does have failing tests, but NOT the one that was updated :) 

Approved!

PR #757, opened 2022-04-25
Oy2 17934

bflynn-collabralink:
Endpoint: https://d2wlhjbrddggem.cloudfront.net/

### Details

Upgrade to serverless v3

### Changes

### Test Plan

No functional changes, just needs full regression.


kristin-at-theta (approved):
Does the documentation for running local need to be updated as well?

Things look good and application deploys and runs!
Approved!

bflynn-collabralink:
> Does the documentation for running local need to be updated as well?
> 
> Things look good and application deploys and runs! Approved!

Local commands should be the same, I dont see anything (yet) that would need to be updated documentation wise. IE installing serverless with the current command listed `npm install -g serverless` should install v3 by default.

PR #766, opened 2022-04-28
SEATool outstream update/fix

kristin-at-theta:
Endpoint: https://d2iylezvrmvuf.cloudfront.net/

### Details

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2

### DIFF
--- a/.github/workflows/deploy-support.yml
+++ b/.github/workflows/deploy-support.yml
@@ -5,7 +5,7 @@ on:
     branches:
       # # List of all branches which should control underlying infrastructure.
       - fix-sechub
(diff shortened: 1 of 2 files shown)


bflynn-collabralink (approved):
still seems to be generating infinite invocations of this lambda and logging it....which is better than sending it back into the stream but maybe we are still receiving a backlog of SEATool records? 
Approving under assumption we can shut off the lambda by setting concurrency to 0 if we want to shutdown the constant invocation but I fear letting it make infinite calls.

PR #770, opened 2022-05-02
Oy2 17916

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-17916
Endpoint: https://do5yi0gi5vg6e.cloudfront.net

### Details

Add additional hint text to medicaid spa id entry field.

### Changes

- made hint text a string array in the TransmittalNumber component and to display each hint on a new line
- converted each instance of idHintText to string array

### Test Plan

1. Login as state submitter
2. From the Submission Dashboard select the "New Submission" button and choose to create a medicaid SPA
3. Verify new hint text appears within the "SPA ID" field


kristin-at-theta (approved):
Looks great, tests well!

Package SPA form is not yet created, so hopefully we'll see this when converting (I think we will, it is coded in the correct place so we'll see it)

Please address the codeClimate findings (can just approve them)

Approved!

services/common/index.d.ts
```diff
@@ -1,3 +1,5 @@
+import { FieldHint } from "../ui-src/src/components/FieldHint";
```
kristin-at-theta, line 1: common should have no dependencies on other services...

kristin-at-theta (changes requested):
Looks alright, except for the import of FieldHint from ui-src into common...

I get it, common has too much Front End detail already - that's "on my list" for moving into ui-src when I get to that side of things.... my preference would be to move the FieldHint interface to the common area for now, then do a real split later.

services/common/index.d.ts
bflynn-collabralink: great feedback! ....moved to common with latest commit

kristin-at-theta (approved):

PR #774, opened 2022-05-09
Dependabot fixes

bflynn-collabralink:
### Details

Dependabot dependency updates plus turn on actions in destroy.sh


kristin-at-theta (approved):
Looks good!  

Approved :)

PR #776, opened 2022-05-16
added aria-describedby to transmittal number field hint

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-17652
Endpoint: https://d2ujkad78dz92o.cloudfront.net/

### Details

Allow screen reader to read informational text surrounding the SPA (and other pages) ID on the form submission

### Changes

- added ids to the hint text
- added aria-describedby to the id input field containing all hint text ids

### Test Plan

1. Login as a state submitter
2. Navigate to Submission list and select the New Submission Button
3. Select SPA -> Medicaid SPA
4. Verify using a screen reader that the "What is my SPA ID?" text is read when user tabs to the link.
5. Verify using a screen reader that the hint text between the label and input  is read when user tabs  to the SPA ID input field


kristin-at-theta (approved):
Checked Medicaid SPA, CHIP SPA, and the Waiver Action form - the What is my... link was read correctly for both ID types and the hint text (including the extra bit on the Medicaid SPA) was read when I tabbed into the actual ID field.

Approved!  Though you probably want to wait to merge until I get develop's database cleaned up...

PR #777, opened 2022-05-16
Oy2 17509 tests

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-17509

### Changes

Fixes two tests that were failing after separation. The tests used the Submission forms and the Package Dashboard. Now they will use the Submission Dashboard when using the Submission forms.


kristin-at-theta (approved):
Approved!

PR #784, opened 2022-05-26
OY2-18479: Set static IDs in test automation

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-18479

### Details

Remove Temp Extension from running test specs since failing due to UI changes that need to be addressed and remove all random ID references from steps. Also, hopefully removed or updated all steps that reference the random ID generator. Small test cleanups as well.


kristin-at-theta (approved):
These changes look great, still working on that "reset script" for you...

Approved!

PR #787, opened 2022-05-27
Oy2 18484

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-18484
Endpoint: https://d3teej317hrsw9.cloudfront.net/

### Details

Restore default table sorting by removing declaration of initialState from PortalTable component. This should be set as param during usage.

### Changes

- Removed initialState from PortalTable (this is passed as param by each usage of the component)
- Fixed PortalTable test which was wrong (presumably made to pass based on faulty initialState declaration)
- fixed id used for sort on package list

### Test Plan

1. Login as state submitter
2. Navigate to submission list and ensure table is default sorted on submitted date


kristin-at-theta (approved):
Wow.... I pretty much edited those SAME THINGS - so guess I'll be merging! :) 

but that means I can easily see what you did!  a lot like what I did for the Inactive filter...

Approved!

PR #789, opened 2022-06-01
fix 2 tests!

Valencia2019:
### Details

Fix two tests that were failing when run in parallel!


kristin-at-theta (approved):
Looks great!  Approved :)

PR #792, opened 2022-06-01
OY2-17236: Change Title Name from "Submission" to "Package" on Package Dashboard

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-17236
Endpoint: https://dbd2mytvgv7tb.cloudfront.net/

### Details
The PageTitleBar title of the package list page should read "Package Dashboard"

### Changes
- edited the title text for the page title bar on PackageList
- went through the PackageList page and renamed other things that should say "package" like the classNames for the table

### Test Plan
1. Log in as any user but the CMS Role Approver
2. go to packagelist
3. Verify that page title bar reads "Package Dashboard"

### DIFF
--- a/services/ui-src/src/containers/PackageList.js
+++ b/services/ui-src/src/containers/PackageList.js
@@ -325,7 +325,7 @@ const PackageList = () => {
     []
   );
 
-  const csvExportSubmissions = (
+  const csvExportPackages = (
     <Button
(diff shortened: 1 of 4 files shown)


bflynn-collabralink (approved):
Look at you going the extra step and covering all the var, function, css names too!

PR #793, opened 2022-06-01
input submittedAt in same format as details page

bflynn-collabralink:
Updated admin function to intake date in same format as displayed on submission details page. (though translation to ET will still need to occur)


kristin-at-theta (approved):
Yeah!  Improvements already!

PR #797, opened 2022-06-02
Updated the Status Text in OneMAC and Tests

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-17024
Endpoint: https://d170hyvh7laop4.cloudfront.net/

### Details
Convert status labels to be all verbs, no subjects.

### Changes
- Updated the statuses lookup table in our common service
- updated unit tests and automated tests to use the new labels
- created migrate function to update old data to use the new labels

### Test Plan
1. Login as a package viewing user
2. Verify that the status labels are verb-only per the story AC


bflynn-collabralink (approved):
Looks good! Im assuming the move of the call to updateParent from changeStatusAny to updateComponent was to cast a wider net for future changes?

PR #799, opened 2022-06-06
OY2-16792: Package Views - Needs to be Read-Only For CMS Users

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-16792

### Details
Read-Only users should not be able to take Actions

### Changes
- used userRoleObj.canAccessForms on the Actions column in the Temporary Extension mini-dashboard
- 

### Test Plan
1. Login as a Read-Only user
2. Navigate to packageList
3. Verify Actions column is not on SPA or Waivers Tabs
4. Select a Package to see it's details
5. Verify the Package Actions column is not there
6. Select a Package with a Temporary Extension child
7. Navigate to the temporary extension mini-dashboard
8. Verify there is no Actions column in the table
9. Test step 2


bflynn-collabralink (approved):
Looks good! I tested as helpdesk user. Didnt know what other roles are read only. We probably should have specified that in the ACs.

PR #807, opened 2022-06-09
Move Reset IDs to the admin service and add a couple more IDs

kristin-at-theta:
Story: Valencia Request1
Endpoint: https://dfsmidccjm8pl.cloudfront.net/

### Details
Valencia had a couple more IDs to add to the reset Data, so took the opportunity to move the lambda into the admin service in the repo.

### Changes
- move resetData into /services/admin
- added the two new IDs

### Test Plan
1. access the endpoint and verify one of the reset IDs exists in the application (add if not)
2. run resetData lambda
3. Verify the components got deleted


bflynn-collabralink (approved):
looks good! I like getting non app functions into admin!

PR #811, opened 2022-06-10
add workflow to temp extensions

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-16961
Endpoint: https://d3otmxz6odvums.cloudfront.net/

### Details

Add Request Extension to temp extension mini dashboard. After submit of a temp extension, return to the temp extension mini dashboard of the parent (base or renewal waiver).

### Changes

- Added Request Extension button to temp extension mini dashboard
- Updated OneMacForm.tsx to expect location.state.parentId and location.state.parentType
- Updated OneMacForm to replace route landing page placeholders when parentId is set
- More refactoring of submitAny

### Test Plan

1. Login as state submitter
2. Create or navigate to a base waiver details page and the temp extension section on the left nav
3. Verify the button "Request Extension" exists at the top of page 
4. Click the "Request Extension" button
5. Verify you are taken to the temporary extension request form
6. Submit a temporary extension request 
7. Verify after submission you are returned to the parent base waiver temp extension page
8. Verify the success banner appears on the page
9. Verify the newly submitted temp extension request is in the table


services/ui-src/src/page/temporary-extension/TemporaryExtensionForm.tsx
```diff
@@ -23,7 +18,7 @@ const temporaryExtensionFormInfo: OneMACFormConfig = {
   idFormat: idFormat,
   actionsByStatus: Workflow.defaultActionsByStatus,
   raiLink: ROUTES.WAIVER_RAI,
-  landingPage: ONEMAC_ROUTES.PACKAGE_LIST_WAIVER,
+  landingPage: ROUTES.DETAIL + "/parentType/parentId#temp-extension",
```
kristin-at-theta, line 21: I think I understand why you are doing things this way, though I believe we do *have* the location state when this component is called.  But... if we are trying for a "config file/execution file" pattern, this does help keep executable code out of this file...

Anyway, it is worth trying and should not be too terrible to change later if it causes problems :)

services/ui-src/src/utils/PackageApi.js
```diff
@@ -71,10 +71,10 @@ class PackageApi {
       );
       throw new Error("Missing required data or uploads");
     }
-    console.log("componentType: ", data.type);
-    console.log("posting to: ", SUBMIT_API_CALL[data.type]);
+    console.log("componentType: ", componentType);
+    console.log("posting to: ", SUBMIT_API_CALL[componentType]);
```
kristin-at-theta, line 75: At some point, I do envision the SUBMIT_API_CALL being a passed variable stored in the OneMACFormConfig

kristin-at-theta (approved):
Tests correct and code reads well :)

Individual comments are where I thought them!

Approved!

services/app-api/utils/newSubmission.js
```diff
@@ -31,11 +31,14 @@ export default async function newSubmission(newData, config) {
     newData.GSI1pk = `OneMAC#${config.packageGroup}`;
     newData.GSI1sk = pk;
   } else {
-    [newData.parentId, newData.parentType] = config.getParentInfo(
-      newData.componentId
-    );
-    newData.GSI1pk = newData.parentId;
-    newData.GSI1sk = config.componentType;
+    if (newData.parentId) {
+      newData.GSI1pk = newData.parentId;
```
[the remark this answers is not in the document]
bflynn-collabralink: yeah, if its in this branch of logic it would be  because it doesnt have a package group so I thought we should expect a parentId....but if not I also kept the old lookup way in the else using the config.getParentInfo

services/ui-src/src/page/OneMACForm.tsx
```diff
@@ -149,25 +145,34 @@ const OneMACForm: React.FC<{ formConfig: OneMACFormConfig }> = ({
...
     if (!event || !event.target) return;
 
-    let updatedRecord = { ...oneMacFormData }; // You need a new object to be able to update the state
+    let updatedRecord = { ...oneMacFormData } as OneMacFormData; // You need a new object to be able to update the state
 
     updatedRecord[event.target.name as keyof typeof updatedRecord] =
       event.target.value;
 
     setOneMacFormData(updatedRecord);
+    console.log("updatedRecord", updatedRecord);
```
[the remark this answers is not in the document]
bflynn-collabralink: > probably don't want this for final application :)

Oooops! Good catch!

services/ui-src/src/utils/PackageApi.js
bflynn-collabralink: > At some point, I do envision the SUBMIT_API_CALL being a passed variable stored in the OneMACFormConfig

Agreed, i keep wanting to refactor that portion, and it keeps getting usurped by other priorities. :)

PR #814, opened 2022-06-15
OY2-18208 (try 2!): SPA RAI on Dashboard to reference Medicaid vs CHIP 

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-18208
Endpoint: https://d2mo2kmpvhr3v2.cloudfront.net/

Details
This application started with Medicaid SPAs... hence, we used "spa" for a lot of things and they should be medicaidspa to differentiate with chipspa (users are getting confused)

Try 2 realizes that changing Submission View should not happen, even with appropriate migrations, so we are leaving everything as /spa and /sparai in the Submission View and only updating the Package View to /medicaidspa and /medicaidsparai

Changes
- Updated the Type column in the Submission UI to say "Medicaid SPA RAI" for "sparai" entries
- Changed the component Type of the Package View Medicaid SPAs from "spa" to "medicaidspa"
- Changed the component Type of the Package View Medicaid SPA RAI Responses from "sparai" to "medicaidsparai"
- wrote migration script to convert existing package data into new format. This will run on develop and val, but not prod. Expectation is that PROD will be migrated using the change-request table just before launch

Test Plan
1. Login as a submitting user
3. SPA RAI type should now read Medicaid SPA RAI


services/common/routes.js
```diff
@@ -56,12 +56,12 @@ export const TYPE_TO_DETAIL_ROUTE = {
...
   [ONEMAC_TYPE.MEDICAID_SPA]: ONEMAC_ROUTES.MEDICAID_SPA_DETAIL,
-  [ONEMAC_TYPE.SPA_RAI]: "",
+  [ONEMAC_TYPE.MEDICAID_SPA_RAI]: "",
   [ONEMAC_TYPE.WAIVER]: ONEMAC_ROUTES.BASE_WAIVER_DETAIL,
   [ONEMAC_TYPE.WAIVER_RAI]: "",
   [ONEMAC_TYPE.WAIVER_BASE]: ONEMAC_ROUTES.BASE_WAIVER_DETAIL,
   [ONEMAC_TYPE.WAIVER_RENEWAL]: "",
   [ONEMAC_TYPE.WAIVER_AMENDMENT]: ONEMAC_ROUTES.WAIVER_AMENDMENT_DETAIL,
-  [ONEMAC_TYPE.WAIVER_EXTENSION]: ONEMAC_ROUTES.TEMPORARY_EXTENSION_DETAIL,
+  [ONEMAC_TYPE.WAIVER_EXTENSION]: "",
```
bflynn-collabralink, line 65: 👍

services/ui-src/src/page/chip-spa/CHIPSPADetail.test.js
```diff
@@ -19,7 +19,7 @@ jest.mock("react-router-dom", () => ({
   ...jest.requireActual("react-router-dom"), // use actual for all non-hook parts
   useParams: () => {
     return {
-      componentType: "spa",
+      componentType: "chipspa",
```
bflynn-collabralink, line 22: good catch!

PR #818, opened 2022-06-17
Oy2 16334 - Update Waiver read-only form: Temporary Extension details

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-16334
Endpoint: https://d28sc16wuq3nda.cloudfront.net/

### Details

As a State User of OneMAC, I want the ability to see the Temporary Extension information of waiver packages so that all waiver family information is organized and easily accessible

### Changes

- Added link from temp extension section mini dashboard to the temp ext details page
- activated routing for temp extension details page
- configured temp extension details page to show only desired elements
- refactored how pageConfig objects are structured so that nav items are commonly declared
- refactored details view to remove each section navigation to its own file

### Implementation Notes

- Some of my refactoring may collide with ongoing refactor work for popups

### Test Plan

1. Login as statesubmitter and navigate to package dashboard
2. Locate or create a base waiver that has an existing temporary extension
3. Click on the temporary extension from the package dashboard 
4. Verify you are navigated to the temporary extension details page
5. Verify temporary extension details page has the components as listed in ACs
6. Navigate back to package dashboard and click on the parent base waiver then on the left nav Temporary Extension section
7. Verify for each temporary extension on the mini dashboard that there is a clickable link
8. Verify that when clicking a temporary extension link you are navigated to the temporary extension details page


kristin-at-theta (changes requested):
I like the cleanup and the break up of that Details View page!

Change Requested: the Additional Information section is not appearing.  The section title should always be there and if there is no details, there is a "no information at this time" type message that should be showing :) 

I don't see anything else amiss - great work!

bflynn-collabralink:
> I like the cleanup and the break up of that Details View page!
> 
> Change Requested: the Additional Information section is not appearing. The section title should always be there and if there is no details, there is a "no information at this time" type message that should be showing :)
> 
> I don't see anything else amiss - great work!

good catch! I didnt even notice it missing from the temp ext page after my recactor. Fixed!

kristin-at-theta (approved):

PR #824, opened 2022-06-22
fix typo in nav header

bflynn-collabralink:
Fixed a Typo in the details view left nav header. PackageOverview => Package Overview


kristin-at-theta (approved):
Nice use of whitespace!  Approved :)

PR #828, opened 2022-06-28
OY2 19015: Reorganize Submission Forms - Text

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19015
Endpoint: https://d36u0zlsw4dios.cloudfront.net/
Figma: https://www.figma.com/file/bpoklF6WABCESBh0DZawDK/State-Users-View-(Refactored)?node-id=6%3A17610

### Details
HCD has reorganized some of the form sections and we'd like the new forms to reflect that.  Also, styles are still being shared between the old and new areas, so breaking those out (and doing a cleanup) is a good idea, too.

### Changes
- moved the h2 to above the paragraph text
- separated the styles from SubmissionForm
- removed extraneous divs, cleaned up some styles
- updated styles to get closer to the Figma

### Test Plan
1. Login as a Submitting User
2. Navigate to each package form
3. Verify that the title has moved
4. Verify that the form looks like the Figma
5. Optional: Navigate to each submission form and verify that nothing looks weird


bflynn-collabralink (approved):
Looks good! Nice job on the css cleanup, I like how you sectioned off the css under  .onemac-form!

PR #831, opened 2022-06-30
Cleanup how Actions are structured for greater clarity

kristin-at-theta:
Story: None, general refactor
Endpoint: https://d1vbd8aai9aev3.cloudfront.net/

### Details
The code for the various Package actions was sprinkled throughout the UI and even duplicated in several places.  This refactor consolidates the action logic and divides actions into distinct components that can be included wherever the action is desired.

### Changes
- Separates each package action into its own component containing all business logic needed to perform that specific action on the "calling package"
- created a mapping between the action list / package / package status arrays and the resulting components
- updated the elements and styles to match UI expectations

### Test Plan
1. Login as a submitting user
2. Navigate to Packages page
3. Use the popup to perform "Withdraw," "Respond to RAI," and "Request Temporary Extension" actions
4. Perform the same actions using the Package Action links in the Package Details pages.


ZacharyRizer (approved):
Anytime we can reduce repeated code the better IMHO. I'm not 100% sure it was necessary to make individual components for each action, but considering there may be more actions down the line, it makes it more extensible and future proof. Other option would be to have dynamic properties on a single 'Action' component, but again, this could get cluttered as more actions are added with specific aspects to each action.

services/ui-src/src/page/action/RespondToRAI.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { Link } from "react-router-dom";
+
+import { ONEMAC_ROUTES, Workflow } from "cmscommonlib";
+
+const RAILink = {
+  [Workflow.ONEMAC_TYPE.MEDICAID_SPA]: ONEMAC_ROUTES.MEDICAID_SPA_RAI,
```
bflynn-collabralink, line 7: I like that we define this here instead of needing TYPE_TO_RAI_ROUTE in common. Can we delete the definition from routes.js in common since it no longer appears to be needed there?

services/ui-src/src/libs/actionLib.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { Workflow } from "cmscommonlib";
+import RequestTemporaryExtension from "../page/action/RequestTemporaryExtension";
+import RespondToRAI from "../page/action/RespondToRAI";
+import Withdraw from "../page/action/Withdraw";
+
+export const actionComponent = {
```
bflynn-collabralink, line 7: I do like how each action is its own component and that could give us flexibility in the future to handle each differently. In fact i think its brilliant the way you made those components the link portion only such that it can be reused on both package and details view. It might however IMO be even better/cleaner to make actionComponent itself a component rather than an array of components and that parent action component can decide which child component to render based on its inputs. Just a thought.

bflynn-collabralink (approved):
I love the consolidation into components such that it makes it possible to use the same thing on both Details and Package View! Made a few suggestions that dont need to be addressed....just considerations.
Approved!

PR #845, opened 2022-07-10
oy2-19299: changed ID format and related text for App K waivers

ZacharyRizer:
Story: https://qmacbis.atlassian.net/browse/OY2-19299
Endpoint: https://d34gw6v64dy81e.cloudfront.net/

### Details

As a OneMAC State User, I need the approved Appendix K ID format to be explained and validated upon submission so that I understand how to correctly submit an Appendix K, which will streamline adjudication.R.

### Changes

- For App K Waivers: changed instructions (subtext that notates correct format), field label hint, Validation Error Message, and FAQ section to reflect this new/correct ID formats (SS-#####.R##.##, SS-####.R##.##)

### DIFF
--- a/services/common/changeRequest.js
+++ b/services/common/changeRequest.js
@@ -318,19 +318,19 @@ export const CONFIG = {
     transmittalNumber: {
       idType: "waiverappk",
       idLabel: "Waiver Number",
-      idFAQLink: ROUTES.FAQ_WAIVER_ID,
+      idFAQLink: ROUTES.FAQ_WAIVER_APP_K_ID,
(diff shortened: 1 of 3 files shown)


kristin-at-theta (approved):
Looks great, liked that you fixed the FAQ link going to the wrong question/answer!

I did find a gotcha here - the existence regex won't work when I update the other ID formats... in fact, it *can't* work unless we make some assumptions on the ID formats... which we are trying to stop doing ... so going to have to go back to Product with that one.  Yay.

Approved!

PR #846, opened 2022-07-12
OY2-19498: Update Waiver ID Format for Base Waivers

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19498
Endpoint: https://d1ocm30airx1rx.cloudfront.net

### Details
Updating the "new" action on the Waiver Action Form to use the new Base Waiver format and the many text changes associated with this.

### Changes
- Added new FAQ section for 1915(b) Base Waiver IDs
- Updated Transmittal Number Component: hint text
- Updated Transmittal Number Component: error text
- Updated Transmittal Number Component: FAQ Link

### Implementation Notes
- also performed npm audit to take care of the dependabot findings
- Did NOT update any lookups of Base Waivers for the Renewals or Amendments, figured it would be easier done in the stories for those components.

### Test Plan
1. Login as a submitting user
3. Click "New Submission->Waiver Action->Waiver Action"
4. Select "New waiver" as the Action Type
5. Verify the hint text is correct
6. Verify the link goes to the correct FAQ question
7. Verify the error message is correct when the format is incorrect
8. Verify the form notices a duplicate ID with a RED error and prevents submission
9. Verify that you can submit the form when all conditions are met.
10. Verify that you can see the submission on the Submission Dashboard and the ID is correct.


bflynn-collabralink (approved):
Looks good! I was going to ask why all your package lock files changed.....but then remembered you did the dependabot merges too!

PR #848, opened 2022-07-14
Oy2 19305

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-19305

### Details

Change Temp Extension ID required format and update validations

### Changes

- Updated form config to reflect new format
- Added additional warning validation to verify parent id exists; including FE and BE; verified new warning appears in email to CMS
- fixed TE bug that allowed TE on any waiver from the mini dashboard

### Implementation Notes

- I attempted using the idExistValidation.existenceRegex to validate the parent id of a temp extension exists however I ran into complications given that we would need special handling to remove the TE portion and append .00 in order to arrive at the parent format. I decided rather than adding in more logic around special handling for the regex to just use the parentid we are already deriving from the getParent function in the config and skip using the regex for this use case. Please let me know if you see a better way!

### Test Plan

1. Test step 1
2. Test step 2


services/ui-src/src/page/OneMACForm.tsx
```diff
@@ -205,10 +205,16 @@ const OneMACForm: React.FC<{ formConfig: OneMACFormConfig }> = ({
...
             let checkingNumber = oneMacFormData.componentId;
-
+            if (
+              idExistValidation.validateParentId &&
+              typeof formConfig.getParentInfo == "function"
+            ) {
+              [checkingNumber] = formConfig.getParentInfo(
+                oneMacFormData.componentId
+              );
+            }
```
bflynn-collabralink, line 215: I attempted using the idExistValidation.existenceRegex to validate the parent id of a temp extension exists however I ran into complications given that we would need special handling to remove the TE portion and append .00 in order to arrive at the parent format. I decided rather than adding in more logic around special handling for the regex to just use the parentid we are already deriving from the getParentInfo function in the config and skip using the regex for this use case. Please let me know if you see a better way!

services/ui-src/src/page/section/TemporaryExtensionSection.tsx
```diff
@@ -124,19 +126,24 @@ export const TemporaryExtensionSection: FC<{
...
-          }}
-        >
-          <Button id="new-temp-ext-button" variation="primary">
-            Request Extension
-          </Button>
-        </Link>
+        {userRoleObj.canAccessForms &&
+          pageConfig.actionsByStatus[detail.currentStatus]?.includes(
+            Workflow.PACKAGE_ACTION.REQUEST_TEMPORARY_EXTENSION
+          ) && (
```
bflynn-collabralink, line 132: Found a bug that allowed TE to be submitted via the TE mini dashboard, therefore tightened up the logic to allow the button under same circumstances as the link in the details section

services/app-api/form/submitAny.js
```diff
@@ -90,8 +93,9 @@ export const submitAny = async (event, config) => {
 
     const results = await Promise.all(promises);
     console.log("ID Matching results are: ", results);
-    if (results[0] != undefined)
-      results.forEach((errorMsg) => {
+    results
+      .filter((errorMsg) => errorMsg != undefined)
+      .forEach((errorMsg) => {
```
bflynn-collabralink, line 98: I was seeing an odd bug whereby the first result was undefined and second result was an error in the array, therefore updated this to filter out all undefined and process rest rather than just look at first item in array

services/app-api/one-seed.json
```diff
@@ -4720,17 +4720,17 @@
         "url": "https://uploads-develop-attachments-116229642442.s3.us-east-1.amazonaws.com/protected/us-east-1%3A86a190fe-b195-42bf-9685-9761bf0ff14b/1655844289419/Attachment%203.1-A%2C%20%234b%2C%20Page%203f%20Track.pdf"
       }
     ],
-    "componentId": "MD.5533.R00.00",
+    "componentId": "MD-5533.R00.00",
```
bflynn-collabralink, line 4723: I only changed this single base waiver format as it is the only approved status waiver whereby you can add a temp extension. The parent waiver needed to be same format as required for TE in order for things to function 100% properly. I assume other base waiver formats will be updated when we get to the base waiver for package view id update story.

kristin-at-theta: That sounds to me like the right approach... for package view, you can assume a conversion from the submission data that can/will include ID updates

services/ui-src/src/page/section/TemporaryExtensionSection.tsx
kristin-at-theta: Very nice find - I would be reluctant to call it a bug, since I'm pretty sure there are no ACs around the button on the mini-dashboard being there or not there based on package status or user permissions.... but I agree that it is the right thing to do (might want to bring it up to product though, or mark it in the Figma?)

bflynn-collabralink: True, more of a possible requirements miss than a bug. I do plan on bringing this up to Ryan/Micha/Jaime.

kristin-at-theta (commented):
The "What is my Waiver Number?" link goes to the old FAQ section for Waiver IDs (missed requirement)

Hmm.... something is off in the format checking for idRegex... because MD-5533.R00-TE01 worked as a "valid" format??  so did MD-5533.R002TE01, but not if you remo.... oh... '.' means match any single character! I think we need to surround '.' with [.] like we did in other places (Probably not a bug you introduced... I had been wondering why we put in those [] chars around the '.'!). Mind fixing that??

kristin-at-theta (approved):
Ok, next time I am just going to "Request Changes" because that was some weird hoops to go through to get GitHub to let me review again (or maybe it is Friday-brain).  

Anyway, saw your fix(es), tested, and it looks good (oh... and I would not have seen it, either, except that I saw the error message disappear after I made a typo!)

Approved!

PR #851, opened 2022-07-18
OY2-19504: Update Waiver ID Format for Waiver Renewals

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19504
Endpoint: https://dilbqymuobcbi.cloudfront.net/

### Details
Update the 1915(b) Waiver Renewal ID format to the new requirements.  Update the supporting texts and validation to handle the new format. 

### Changes
- updated the Waiver Renewal ID regex for format matching
- updated the field hint to the new format
- updated the id regex to the new format
- updated the error message to the new format
- updated the FAQ page with a new section for 1915(b) Waiver Renewals

### Implementation Notes
- original existence regexes assumed that the parent ID was a subset of the child component ID.  This is no longer the case, so had to refactor.

### Test Plan
1. Login as a submitter user
2. Navigate to "Submission Dashboard" (Dashboard link on nav)
3. Click "New Submission->Waiver Action->Waiver Action"
4. Select "Request for waiver renewal" from Action Type dropdown
5. Verify the field hint format matches the AC
6. type in an incorrect number and verify the error formats match the AC
7. Click the "What is my Waiver Number?" link and verify the correct FAQ section is shown and matches the AC
8. Verify that if the base number does not have a submission, form shows a blue message and you can submit
9. Verify that if a duplicate Renewal number is entered, a red error message is shown and you cannot submit
10. Verify the submission CMS notice email contains the blue warning message

### DIFF
--- a/services/common/changeRequest.js
+++ b/services/common/changeRequest.js
@@ -283,22 +283,23 @@ export const CONFIG = {
     renewalTransmittalNumber: {
       ...waiverBaseTransmittalNumber,
       idFieldHint: [
-        { text: "Must follow the format SS.####.R## or SS.#####.R##" },
+        { text: "Must follow the format SS-####.R##.00 or SS-#####.R##.00" },
       ],
-      idFormat: "SS.####.R## or SS.#####.R##",
-      idRegex: "^[A-Z]{2}[.][0-9]{4,5}[.]R[0-9]{2}$",
+      idFormat: "SS-####.R##.00 or SS-#####.R##.00",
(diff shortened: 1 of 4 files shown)


bflynn-collabralink (approved):
Looks good! This is almost the solution i went with for package view until i realized we already had the parentId as part of the config and it didnt make sense to recreate that via regex. I think it makes sense though to not try to port all that parent logic over to submission view.
Didn't see anything off during manual testing.
Approved!

PR #852, opened 2022-07-18
OY2 19153: Relax RAI Response ID Format Validation

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19153
Endpoint: https://d2hrowy4zchy95.cloudfront.net/

### Details
It was decided to remove the formatting rules for RAI responses.  Now we should only check if the submitter can submit for the territory (first two characters of ID) and if the ID exists in our data.  Note that this is in submission view, so we don't know the types of the ids, just that they match.

### Changes
- remove the id formatting on the RAI forms
- update the text accordingly
- update the id checking to be a blue warning message when the id is not found

### Test Plan
1. Login as a submitter user
2. Navigate to the Dashboard (Submission)
3. Note several IDs of various types
4. Click "New Submission->State Plan Amendment->Respond to Formal Medicaid SPA RAI"
5. Note that you get an error until you have entered a valid territory code
6. Note that you get a blue warning message for typing anything that does not match an ID
7. Note that typing in an ID from the Dashboard (any type) clears the errors and warnings
8. Verify that the Submit button is disabled with a red message, but enabled for the blue and no messages
9. Submitting correctly creates the Response to RAI Submission
10. also check the "Respond to Formal CHIP SPA RAI" form
11. also check the "Respond to Waiver RAI" form

### DIFF
--- a/services/app-api/changeRequest/CHIPSPARAI.js
+++ b/services/app-api/changeRequest/CHIPSPARAI.js
@@ -13,26 +13,7 @@ class CHIPSPARAI {
    * @returns {String} any errors
    */
   async fieldsValid(data) {
-    let areFieldsValid = false;
-    let whyNot = "";
-    let doesExist = false;
-
-    try {
-      doesExist = await packageExists(data.transmittalNumber);
-    } catch (error) {
-      console.log("CHIPSPARAI packageExists call error: ", error);
-      throw error;
-    }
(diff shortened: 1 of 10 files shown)


bflynn-collabralink (approved):
Looks good according to the ACs. We better get our prod id update scripts warmed up!

PR #853, opened 2022-07-18
OY2-19588 - batchUpdateChangeRequestId implementation

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-19588
Endpoint: https://d1git28twbsi47.cloudfront.net/

### Details

Create an admin function that can take a csv file and perform updates based on the given inputs.
Required data points: Old ID Format, New ID Format, Type, State

### Changes

- Created batchUpdateChangeRequestId admin function

### Implementation Notes

- I added a mapping in this function for Type given by business for "Medicaid Spa" -> "spa"; future stories for other types may want to add on new mappings

### Test Plan

1. Identify a medicaid spa change request to modify
2. Use the CSV format as found when exporting the given excel to CSV (Also see handlers/batchUpdateChangeRequest.json)
3. Open the test tab for the lambda in aws console and paste the csv content into the event json
4. Run
5. Verify in the log output which data successfully passed and which did not get updated
6. Verify in the database that updates were made

### DIFF
--- a/services/admin/README.md
+++ b/services/admin/README.md
@@ -12,6 +12,11 @@ Run this lambda to delete the components used in the automated testing. Right no
 
 Update a given change request transmittalNumber based on its current transmittalNumber, type, and submittedAt timestamp and prepend a message onto the addtionalInfo
 
+#### batchUpdateChangeRequestId
+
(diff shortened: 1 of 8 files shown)


kristin-at-theta (approved):
Looks great - I liked that you documented the process - though it still took me four tries to get it right :P   but a kind error out is better than doing the wrong thing!

Anyway, looks good and worked - approved!

PR #855, opened 2022-07-20
FIX 19504: submit button concurrency bug

kristin-at-theta:

### Details

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2

### DIFF
--- a/services/ui-src/src/changeRequest/SubmissionForm.tsx
+++ b/services/ui-src/src/changeRequest/SubmissionForm.tsx
@@ -292,29 +292,34 @@ export const SubmissionForm: React.FC<{
         displayMessage = formatMessage;
         setTransmittalNumberStatusMessage(displayMessage);
       }
-
-      let formReady = false;
-      if (
-        (!formInfo.actionType || changeRequest.actionType) &&
-        (!formInfo.waiverAuthority || changeRequest.waiverAuthority) &&
-        (displayMessage.statusLevel === "warn" ||
-          !displayMessage.statusMessage) &&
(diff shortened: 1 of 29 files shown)


services/ui-src/src/changeRequest/SubmissionForm.tsx
```diff
@@ -292,31 +292,36 @@ export const SubmissionForm: React.FC<{
...
+      formReady = true;
+
+    setIsSubmissionReady(formReady);
+  }, [
+    areUploadsReady,
+    changeRequest,
+    formInfo,
+    transmittalNumberStatusMessage,
+  ]);
+
```
bflynn-collabralink, line 324: I like having more targeted useEffect blocks rather than giant ones. Was this an issue in this case because the `formReady` logic was occuring after the promises which may or may not have been resolved by the time the logic looked at the `displayMessage` property?

kristin-at-theta: I did not do stack traces :) but to the best of my knowledge and observations, you are exactly right, the formReady code was happening before the Promise returned... so, there was no error message at that time, so the form was ready.

I agree about the more targeted useEffects, too... and I *thought* that was how the form was, but clearly it is not... I looked in the history some, and did see that it was originally separated, but somewhere the code got consolidated... I would have to do a more thorough review to see exactly when and really, that is only important if we think it was done for a Reason....

bflynn-collabralink (approved):
I couldnt get it to break. Approved!

PR #856, opened 2022-07-20
update format tests in submission view

Valencia2019:
### Details

Main thing to note: we are now checking for the submit button being disabled/enabled after putting in invalid and valid IDs, respectively.

### Changes

- Ended up finding some other issues. Updated some of it. Got down a rabbit hole and decided to wrap up where I am.


kristin-at-theta (approved):
Yes, the logic looks much more useful for finding errors in our validation logic!

Hopefully, they won't change said logic too often so we don't have to update a lot!

Anyway, Approved :)

PR #865, opened 2022-07-22
OY2 17913: BUG-Skip To Main Content - Hash Router

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-17913
Endpoint: https://dgu2m8zuq0hfl.cloudfront.net/

### Details
508 compliance requires a "Skip to Content" link at the top of the page to provide a path directly to that page's individual content.  This link was not working on the Package View Details pages.

### Changes
- Added #main as an allowed anchor on the details page tab list
- set #main to be the same content as #component-detail (the default setting)

### Test Plan
1. Login as a user who can view details pages
3. Select any component details by clicking an ID
4. Use the "Skip to Content" feature... one way to do this is to click in the URL bar of the browser and hit tab.
5. Observe that the page has jumped to the content, which should be the Package Overview->Package Details section


bflynn-collabralink (changes requested):
This seems to work, however if on a different section such as Additonal Information or the Temporary Extensions. In those cases, clicking the Skip to Content redirects back to the details view.

bflynn-collabralink (approved):
After discussions on implementation paths, we have agreed to go this route for now.

Approved!

PR #866, opened 2022-07-22
remove the character counter and relevant styles

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19662

### Details
The character count feature in the Additional Information text area was triggering for every character typed and reading aloud the character count and the character limit.  This is very distracting and it was decided that the character counter should be removed (though the character limit is still 4000 chars)

### Changes
- Removed the div containing the character count and limit
- updated the style sheet to remove the styles associated with that div
- this was done for ALL forms, both Submission and Package

### Test Plan
1. Login as a submitter user
2. Navigate to Dashboard (Submission)
3. Click "New Submission" and travel to a new form
4. turn on a screen reader
5. type in the Additional information box
6. Verify it says what letters you are typing, but does not tell you how many characters are left til the limit
7. Do the same for the Packages forms

### DIFF
--- a/services/ui-src/src/changeRequest/SubmissionForm.tsx
+++ b/services/ui-src/src/changeRequest/SubmissionForm.tsx
@@ -457,9 +457,6 @@ export const SubmissionForm: React.FC<{
               value={changeRequest.summary}
               maxLength={config.MAX_ADDITIONAL_INFO_LENGTH}
             ></TextField>
-            <div className="char-count">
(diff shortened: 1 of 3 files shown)


bflynn-collabralink (approved):
I like that you removed the no longer needed css too! Im not sure I would have thought of that right away. Confirmed screen reader still reads text typed in the text area.
Looks good! Approved!

PR #867, opened 2022-07-22
Oy2 19634 - base waiver id format for package view

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-19634
Endpoint: https://d1es38x71h06ou.cloudfront.net/

### Details

Update base waiver id required format for package view

### Changes

- Updated baseWaiver and BaseWaiverForm config for format and FAQ link

### Implementation Notes

- Tested to ensure TE on base waiver still functions/links property; Renewal/Waiver not yet implemented.

### Test Plan

1. Login as state submitter
2. Navigate to package list and click New Submisssion and select Waiver Action -> Base Waiver
3. Enter a waiver authority, proposed effective date, and add any attachment
4. Enter a waiver id that already exists and verify red error message (form is not submittable)
5. Enter a waiver id in the incorrect format and verify red error message (form is not submittable)
6. Enter a waiver id in the correct format and submit


kristin-at-theta (approved):
Looks good, tests well, though I found something odd that is also in develop.... as in, I can get a red error message on the ID and still click the Submit button.  The back end for packages *also* checks, so it fortunately fails in actually making the duplicate entry (was not the case for Submission View)

Approved!

bflynn-collabralink:
> 

Interesting, i thought the submit was blocked but yeah it appears on duplicate id submit it allows and fails on back end. Maybe ill dig into this today even though it wont block the story. Thanks!

kristin-at-theta (approved):
Thanks for fixing that submit button!  Changes approved :)

PR #868, opened 2022-07-26
OY2-19795: Onemac - Security Hub Findings 2022-07-06 Report Review and Triage

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19795
Endpoint: https://d1es38x71h06ou.cloudfront.net/

### Details
Some security hub findings can be solved for future deployments through modifications to the serverless.yml.

### Changes
- added the bucket helper to the admin service - this runs a script to automagically set the public access to the s3 bucket to denied for all types of public access.  The account already has this setting as a default on new buckets, but sec hub wants it to also be assigned per bucket.
- allowed X-Ray tracing to be used in the lambdas of app-api service.  Not sure why this is important or how we would want to use them (and frankly, I feel like this Decreases security, but I am not an expert), but they are added.

### Test Plan
1. Deployment should be successful
2. When looking at a lambda function, should see that X-Ray tracing is a Resource in "Configurations->Permissions"
3. When looking at the s3 bucket for the admin service, should see "Block All Public Access" is set to ON in the Permissions


bflynn-collabralink (approved):
Looks ok to me. Verified xray traces are showing up in cloud watch for api / lambda calls.
I did notice a difference in the url used for the serverless-s3-bucket when compared to usage in the root onemac/package.json but not sure if that matters or is intentional.
Approved.

Dark-Knight-000 (approved):
Hmmm. I thought I added the bucket helper previously. I must have missed a spot.

PR #870, opened 2022-07-26
Bug: blue messages should not get priority over red

kristin-at-theta:
Story: BUG found in a meeting about IDs
Endpoint: https://d1w47fkcenmnej.cloudfront.net/

### Details
While going over the Waiver Renewal ID validations, we noticed that the blue "waiver not found" message was being shown when the full number was a duplicate (and the parent was not findable), this condition should show the red "duplicate ID" message first... basically, all red messages should have priority over all blue messages.

### Changes
- split messages array into red and blue
- updated logic so that red shows first, then blue (note, format messages are checked outside of the the exist messages)

### Test Plan
1. Login as a submitting user
2. Navigate to "Dashboard->New Submission->Waiver Action->Waiver Action"
3. Select "renewal" action
4. try to use an id that exists in the dashboard as a renewal, but does not have a base waiver
5. Verify that the red message is shown

### DIFF
--- a/services/ui-src/src/changeRequest/SubmissionForm.tsx
+++ b/services/ui-src/src/changeRequest/SubmissionForm.tsx
@@ -217,7 +217,8 @@ export const SubmissionForm: React.FC<{
       warningMessageCode: "",
     };
 
-    let existMessages: Message[] = [];
+    let redMessages: Message[] = [];
+    let blueMessages: Message[] = [];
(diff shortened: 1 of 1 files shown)


bflynn-collabralink (approved):
Looks good to me. Was not able to submit a duplicate renewal or amendment with no base.
Approved!

PR #872, opened 2022-07-28
update admin function to allow waiver, TE, and do test runs

bflynn-collabralink:
Update to admin function to allow test event via event property such that no updates occur but anticipated changes are output.


ZacharyRizer (approved):
Test runs and outputs are always good ideas to me!

PR #873, opened 2022-07-28
oy2-19905: script was updated for batchUpdateChange... and update to …

ZacharyRizer:
Story: https://qmacbis.atlassian.net/browse/OY2-19905
Endpoint: https://dinuj2q42nn3g.cloudfront.net/

### Details

As a MACPRO developer, I need the historical OneMAC Waiver entries to have IDs to have a dash rather than a dot, so that the entries in OneMAC match their SEA Tool counterparts.

### Changes

- Added to the type options in 'batchUpdateChangeRequestId.js'
```
          case "1915(c) Appendix K Amendment":
            return "waiverappk";
```

### Implementation Notes

- Tested and ran the lambda to update the ID's


bflynn-collabralink (approved):
Looks good! Thanks for adding the deploy step to the readme!

kristin-at-theta (approved):
I agree, nice to have the READMEs be useful!  I do read them :) 

Approved!

PR #874, opened 2022-07-29
OY2 19345: Update Temporary Extension ID Format - Submission View

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19345
Endpoint: https://db78kopwj2al9.cloudfront.net

### Details
Temporary Extension Format is getting re-established.  No duplicates are allowed, and parent IDs are not checked.  This Pull Request updates the hint text, the format error message, the format regex, the FAQ, and the existence checks.

### Changes
- new format is SS-####(#).R##.TE## (R00 and TE00 are allowed values)
- in common/changeRequest, updated the details used for transmittalNumber for CONFIG
- Field Hint has new text with new format, broken into two lines because the max-width of 460px is causing an ugly break in the middle of the ID. (SS- )
- error message is updated with new format
- existence validation is updated to prevent duplicates (red error)

### Test Plan
1. Login as state submitter user
2. Navigate to Dashboard
3. Click New Submission->Waiver Action->Request Temporary Extension
4. Verify the hint text is correct
5. Verify the FAQ link goes to Temporary Extension specific section and it reads correctly per the AC
6. Type an ID that does not match the format
7. Verify Submit is disallowed and a red error message is presented
8. Type a valid format ID, red message should go away
9. Submit that extension, noting the ID used
10. Click New Submission->Waiver Action->Request Temporary Extension
11. Attempt to use the same ID
12. Verify the duplication error message is red and you cannot submit form


bflynn-collabralink (approved):
Looks good!
The FAQ section was placed where the AC says but I do find it odd to have the all the other ID format questions come first then the Attachment questions, but this one got stuck in the middle.
Approved.

ZacharyRizer (approved):
All looks correct to me!

PR #879, opened 2022-08-03
Oy2 19876 v3

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-19876
Endpoint: https://d1cgarra2aoa75.cloudfront.net/

### Details

Move (most) github secrets to aws ssm parameter store

### Changes

- added branch specific ssm parameters for all email based params
- updated serverless config to pull vars from ssm

### Implementation Notes

- Github secrets were not yet deleted since they are actively used until this gets up to PROD
- Initially I had also created /configuration/default secrets however that could not be deployed across multiple branches in the same aws environment since they are the same ssm param for each branch. Could have done some conditional logic but I figured since each branch will have its own param anyway, what is the point of even having a default?

### Test Plan

No functional changes. Just ensure application works as it was, especially regarding emails.


README.md
```diff
@@ -13,10 +13,6 @@ An official submission system for email-based state plan amendments (SPAs) and s
 The following environment variables can be set to change the configuration of the application (reference [build_vars.sh](./.github/build_vars.sh)):
 Lengend: R = Required, O = Optional
```
kristin-at-theta, line 15: You are always so good about this part of the job!  Kudos!

services/app-api/resources/base.yml
```diff
@@ -3,4 +3,4 @@ Conditions:
     Fn::Not:
       - Fn::Equals:
           - ""
-          - ${env:IAM_PERMISSIONS_BOUNDARY_POLICY, ""}
+          - ${self:custom.iamPermissionsBoundaryPolicy, ""}
```
kristin-at-theta, line 6: nice cleanup here as well

services/app-api/serverless.yml
```diff
@@ -21,15 +21,14 @@ custom:
...
-  emailSource: ${opt:emailSource, env:CMS_SPA_FORM_FROM_EMAIL, "spa-reply@cms.hhs.gov"}
-  userAccessEmailSource: ${opt:userAccessEmailSource, env:CMS_USER_ACCESS_FROM_EMAIL, "OneMACAccess@cms.hhs.gov"}
-  reviewerEmail: ${opt:reviewTeamEmail, env:CMS_SPA_FORM_CMS_EMAIL, "OneMAC@cms.hhs.gov"}
-  reviewerCHIPEmail: ${opt:reviewTeamCHIPEmail, env:CMS_CHIP_FORM_CMS_EMAIL, "OneMAC@cms.hhs.gov"}
-  chipCcEmail: ${opt:chipCcEmail, env:CMS_CHIP_FORM_CC_EMAIL, ""}
-  systemAdminEmail: ${opt:systemAdminEmail, env:CMS_SYSTEM_ADMIN_EMAIL, ""}
-  productionNoEmailDebug: ${opt:productionNoEmailDebug, env:NO_EMAIL_DEBUG, "OneMAC@cms.hhs.gov"}
-  testingEmail: ${opt:testingEmail, env:TESTING_TEAM_EMAIL, ""}
+  iamPath: ${ssm:/configuration/${self:custom.stage}/iam/path, ssm:/configuration/default/iam/path, "/"}
+  iamPermissionsBoundaryPolicy: ${ssm:/configuration/${self:custom.stage}/iam/permissionsBoundaryPolicy, ssm:/configuration/default/iam/permissionsBoundaryPolicy, ""}
```
kristin-at-theta, line 25: I think it is the right call to remove the opt ability from these variables... not sure we want to allow that kind of variability in a deployment

Also.... I think we have had trouble in the past when the dev ops team updated the ssm:/configuration/default/iam/permissionsBoundaryPolicy and did not update the secret as well... so this should help with those sorts of troubles, too!

services/app-api/serverless.yml
```diff
@@ -21,15 +21,14 @@ custom:
...
+  emailSource: ${ssm:/configuration/${self:custom.stage}/email/cms_spa_form_from_email, ssm:/configuration/default/email/cms_spa_form_from_email, file(resources/ssm-params.yml):Resources.CmsSpaFormFromEmailDefault.Properties.Value}
```
kristin-at-theta, line 26: I also like how you are referencing the same default as in the ssm-params.yml

services/app-api/resources/ssm-params.yml
```diff
@@ -0,0 +1,40 @@
...
+
+  CmsSpaFormChipEmail:
+    Type: AWS::SSM::Parameter
+    Properties:
+      Type: StringList
+      Description: Branch specific email address of recipient for cms chip spa form submission emails
+      Name: !Sub /configuration/${self:custom.stage}/email/cms_spa_form_chip_email
+      Value: "OneMAC@cms.hhs.gov"
+
+  CmsSpaFormChipCcEmail:
```
kristin-at-theta, line 34: looks like this value defaults to "" in previous branches

kristin-at-theta (approved):
I'll admit I don't know exactly what to look for in order to see any issues, but I see the changes and they make sense.  They even clean up some details from before!  

I checked that the application runs, I was able to submit a Medicaid SPA with an attachment, and I looked in the console and the variables for the lambdas (specifically the submit) look the same (except the Chip cc, noted in the comments).

I looked in the account and did not see any errors in the stacks or any New security hub findings for the branch.  So... I think we are GOOD!

I did notice that some of the variables being stored in the ssm for the QuickStart are still in the Actions for this branch, but we have a great pattern now for how to move them, so can do that, as needed!

Approved!

services/app-api/resources/ssm-params.yml
bflynn-collabralink: oh yes, this is a good point I forgot to mention in the PR. As is it would send a duplicate email to that address since it is both the default To: and CC: address, we probably dont care in environments < production and in production we will need to update the CC with the current list of individual email addresses once deployed. I couldnt figure out a way though to get cloudformation to accept an empty or null parameter value via serverless to get it setup so decided to go this route.

PR #881, opened 2022-08-05
Oy2 18704

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-18704
Endpoint: Doesn't matter!

### Details
Added a section to the destroy.yml to send a dispatch to bigmac repo and tell it to delete the topics associated with the local branch.  Also updated the partition variables requested in another story.

### Changes
- added the workflow changes to destroy.yml
- used my own PAT because I believe I have write access to bigmac repo
- added partition variables to stream-functions/serverless.yml

### Implementation Notes
- because all destroys use the develop branch destroy.yml... have to merge first, then can test....


bflynn-collabralink (approved):
I can confirm the topic names appropriately match what we are publishing to in stream-functions service. Otherwise, lets see what happens!

Approved!

PR #883, opened 2022-08-08
OY2-19988 - added ability to append additional error message information per submission type

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-19988
Endpoint: https://d1cwt4cui1vr0t.cloudfront.net/

### Details

Needed to add additiional error information displayed to user for Waiver Amendment and App K Amendments when format is not correct.

### Changes

- Updated TransmittalNumber component to parse the error message for newline characters and render each section in its own div
- Added new field to each submission type in ChangeRequest.js config for the additional error text (including the newline so that it will be displayed on an additional line to prevent long line wrapping.

### Implementation Notes

- While I feel the way I implemented this is semi-hacky (by parsing for newline characters in the TransmittalNumber component); I felt that was better than restructuring the entire error handling mechanism to allow arrays in every submission view form and in the TransmittalNumber component. If this were for package view I probably would have done the restructure of how errors are handled.
- Alternatively I could have just appended the message to the current message for those types and let it wrap but it looked odd because it went far past the input box. 

### Test Plan

1. Login as state submitter
2. Go to Submission Dashboard
3. New Submission -> Waiver Action -> Waiver Action
4. Select Waiver Amendment for Action Type
5. Type only the State portion of the amendment number
6. Verify the error message has the additional error text as described in AC.
7. Verify after typing a valid waiver number that the error disappears.
8. Repeat steps 5-7 for New Submission -> Waiver Action -> Appendix K Amendment

### DIFF
--- a/services/common/changeRequest.js
+++ b/services/common/changeRequest.js
@@ -48,6 +48,9 @@ const waiverBaseTransmittalNumber = {
   idFAQLink: ROUTES.FAQ_WAIVER_ID,
 };
 
+const waiverAddtionalErrorMessage =
+  "\nFor amendments, the last two digits start with “01” and ascends.";
+
 export const CONFIG = {
   [TYPE.CHIP_SPA]: {
     pageTitle: "Submit New CHIP SPA",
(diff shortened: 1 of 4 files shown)


ZacharyRizer (approved):
I agree that refactoring would be a bit overkill for this considering Submission view. And the new line parsing isn't 'too hacky' for me!

PR #884, opened 2022-08-08
OY2 19856: Temporary Extension: Add Text Box and Validation of Parent ID

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19856
Endpoint: https://d1d6c6uho0r6ro.cloudfront.net/

### Details
Product would like to add a Parent ID field to the Temporary Extension Submission form so that it is easier for the email recipient to know which Waiver has the extension request.

### Changes
- updated the Waiver Extension emails to deliver the new numbers and labels
- added the parentWaiver config fields to the formInfo types
- modified the SubmissionForm to manage the two instances of the TransmittalNumber component
- updated the labels and hint text according to the ACs
- fixed a react warning about needing key elements
- updated TransmittalNumber component to have more custom pieces, so two can be put into one page (ids, etc)

### Implementation Notes
- implemented as a second TransmittalNumber component within the form.

### Test Plan
1. Login as a submitting user
2. Navigate to Submission Dashboard (Dashboard link)
3. Copy an ID of any submission type
4. Click "New Submission->Waiver Action->Request Temporary Extension"
5. Verify the look and text of the new field matches the AC
6. Verify that an ID that is not in the database displays an error and submit is disabled (when all else is correct)
7. Verify that an ID in the database has no error and submit is allowed (when all else is correct)
8. Verify that the look and text of the Temporary Extension Request Number matches the AC
9. Submit a valid Temporary Extension Request
10. Navigate to the details page for that Temporary Extension Request
11. Verify that the ACs for the details page are met.


ZacharyRizer (approved):
Everything looks clean to me. I think I lean towards this simpler (albeit maybe slightly more rough) approach of reappropriating the transmittalNumber component (rather than a brand new component).

PR #885, opened 2022-08-08
OY2-19722 & 11100: Idle/Logout Timer implementation

ZacharyRizer:
Story: https://qmacbis.atlassian.net/browse/OY2-19722 & OY2-11100
Endpoint: https://d3ugk3tk6fxahq.cloudfront.net/

### Details

As a user of OneMAC, I want a session time-out warning to show on screen, so that after a duration of 1 hour the session will time-out requiring me to log back in. Warning box will opening once the 15 minute threshold has been breached.

### Changes

- Created a wrapper 'component', IdleTimerWrapper.tsx, that contains all implementation and logic for the timer and returns a blank React fragment
- App.tsx imports and 'renders' the IdleTimerWrapper so that the entire application has access

### Implementation Notes

- The timer is currently only using time to open the dialog, eventually if we wanted it would be simple to incorporate idle triggers (mouse movements, clicks, keyboard, etc) to reset the timer
- It is setup to handle multiple tabs and should trigger the dialog on all tabs open
- It currently displays the amount of time remaining when the dialog opens (usually will show fifteen minutes except in special cases). However this timer is not live, so if the tab is open in the background and the timer message pops up, the time displayed could be stale by the time the user sees it. 
- The App.test.js 'render' test was failing requiring a workaround mock to be added. Still investigating why the IdleTimer package has difficulty with 'render' tests. 

### Test Plan

1. Login as state submitter
2. Wait 45min (sorry)
3. Dialog Box will open stating that the session will expire in about 15 minutes
4. Click continue or outside of box to continue browsing - after another 15 minutes the user will be logged out and redirected to the homepage
5. Clicking logout in the dialog will logout the user and take them back to the home page


kristin-at-theta (approved):
Looks good, I like that you pulled out the logout code into its own lib.

Yup, got the popup after 45 minutes (or so) and it logged me out when I clicked the logout button.  Trying the "other option" now, but I feel like it will work!

I'm not a fan of losing the test coverage, but I can't think of any way to test it, either.

Approved!

PR #886, opened 2022-08-08
Annotation test

bflynn-collabralink:
Story: Brandons made up agenda 01
Endpoint: https://dig84sk6hecy7.cloudfront.net

### Details

I wanted to add an easier way to get to the endpoint created upon deploy without digging thru output logs. Found this mechanism to output annotations to the github actions output.

### Changes

- added new step in deploy to send a notice annotation for the endpoint url

### Test Plan

1. Push code to a branch
2. Verify the Endpoint Url is output as a notice annotation


kristin-at-theta (approved):
Yup, looks like I'd expect!

Approved!

PR #913, opened 2022-08-11
Testing tests

bflynn-collabralink:
Story: Devops
Endpoint: Should we get rid of this from the template now that the bot will post the endpoint?

### Details

Add a jest runner that will add an annotation for each failed unit test

### Changes

- Added github-actions-reporter
- use reporter flags when running npm test

### Test Plan

1. commit a failed test
2. verify the action is annotated with the name and line of the test that failed

Note: you can reference my failed test in previous action deploy run:
https://github.com/CMSgov/onemac/actions/runs/2843033705

### DIFF
--- a/github-actions-reporter
+++ b/github-actions-reporter
@@ -1,0 +1,36 @@
+class GithubActionsReporter {
+  constructor(globalConfig, options) {
+    this._globalConfig = globalConfig;
+    this._options = options;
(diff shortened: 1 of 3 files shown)


kristin-at-theta (approved):
Fan-freakin'-tastic!  I love it :) 

Approved

PR #918, opened 2022-08-16
OY2 19927: Create Submission Form Package View - 1915(b) Waiver Renewal for Initial Waiver not in OneMAC Dashboard

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-19927
Endpoint: see comment

### Details
allow submission of Waiver Renewals for Initial Waivers that did not originate in OneMAC

### Changes
- added the Waiver Renewal form
- added a text field to hold the number of the waiver being renewed.  It looks for the given number and verifies it is for something of the correct type and status

### Test Plan
1. Login as a submitting user
3. Click "New Submission->Waiver Action->Waiver Renewal"
4. Verify that the form appears as described in the AC (though places where the AC says "Amendment" usually should say "Renewal")
5. Verify the Existing Number to Renew is a required field
6. Verify that the only valid values for the Existing... field are Waiver Initial or Renewals with Approved status
7. Verify that submitting the form correctly saves the Waiver Renewal and places it on the dashboard
8. For testing purposes, you probably used a number that is on the dashboard, it should also show as a child, there.


services/ui-src/src/page/waiver-renewal/WaiverRenewalForm.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { FC } from "react";
+import OneMACForm from "../OneMACForm";
+import { defaultOneMACFormConfig, OneMACFormConfig } from "../../libs/formLib";
+import { ROUTES, ONEMAC_ROUTES, waiverRenewal } from "cmscommonlib";
+
+const waiverRenewalIdFormat: string = "SS-####.R##.00 or SS-#####.R##.00";
+
+export const waiverRenewalFormInfo: OneMACFormConfig = {
+  ...waiverRenewal,
+  ...defaultOneMACFormConfig,
```
bflynn-collabralink, line 10: do we want to switch this around to have the default spread first in case anything in the more specific config needs to overwrite something in the default?

bflynn-collabralink (changes requested):
Waiver Renewals seem to be working as desired however something is failing for all other types (spa, chip spa, initial waiver)

services/ui-src/src/page/waiver-renewal/WaiverRenewalForm.tsx
kristin-at-theta: yes, good catch!

bflynn-collabralink (approved):
Looks good to me.

Approved!

PR #928, opened 2022-08-26
OY2 20367: BUG - Spinning circle appears in browser after clicking Submit button for CHIP SPA

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-20367

### Details
User is getting stuck at spinner when attempting to submit a CHIP SPA with several very small attachments.  Initial triage discovered this is the error we have seen before where AWS throttles the uploads and we do not receive the error message.

Apparently Amplify has updated and we now do get the error message returned for the 503!

### Changes
- fixed a bug in the catch that wrongly assumed an error was a string
- updated the error handling to pass back a RESPONSE_CODE so the error gets shown as a red alert (going to be hard to test that, I tested it by forcing a fail condition on the function in the code itself because I ran out of other ideas)
- changed the upload function so it retries an upload 10 times (9?) after receiving the 503 error

### Test Plan
1. Login as a submitting user
2. Navigate to Dashboard
3. Click "New Submission->State Plan Amendment(SPA)->CHIP SPA"
4. use a valid new CHIP SPA id
5. Attach small documents to most of the types (user did all but one, max file size was 1.2MB)
6. Open Inspect
7. Submit the form... if there is an error, verify that the submission was, in fact, successful
8. If no error, you might have to wait a while before retrying as I believe this error may be happening because AWS will send 503 messages to slow down if it decides to add more resources because it takes a moment for those to initialize... there are other reasons, but that one seems most likely to me....


ZacharyRizer (approved):
I went through the testing strategy and saw no errors.

- logged in
- went to the CHIP SPA form
- gave it valid id
- attached small attachments to all sections
- submitted
- no error

services/ui-src/src/utils/s3Uploader.js
```diff
@@ -68,27 +68,38 @@ export async function uploadFile(file) {
...
-    retPromise = Promise.reject(error);
+      retPromise = Promise.resolve(result);
+      numTries = 0;
+    } catch (error) {
+      numTries++;
+      retPromise = Promise.reject(error);
+      if (error.message.indexOf("failed with status code 503") !== -1) {
+        console.log("got 503, numTries: ", numTries);
+      } else {
+        console.log("got a different error: ", error);
```
bflynn-collabralink, line 99: doesnt affect anything but do we need the logic branch here to detect the 503 if in both cases we are just incrementing the try counter and trying again? In both cases we could just log the error object to console (which would include the errror.message).

services/ui-src/src/utils/s3Uploader.js
```diff
@@ -43,19 +45,17 @@ export async function uploadFiles(fileArray) {
...
         })
         .catch((error) => {
-          if (error.indexOf("No credentials") !== -1) {
-            reject("SESSION_EXPIRED");
+          if (error.message.indexOf("No credentials") !== -1) {
+            reject(RESPONSE_CODE.SESSION_EXPIRED);
           } else {
-            console.log("Error uploading.", error);
-            reject("UPLOADS_ERROR");
+            reject(RESPONSE_CODE.UPLOADS_ERROR);
```
bflynn-collabralink, line 51: I like how you cleaned up the literals into their respective constant values even though it wasnt directly part of the code you needed to modify!

services/ui-src/src/utils/s3Uploader.js
kristin-at-theta: HA HA - good eye!  That was leftover from when I was not retrying on anything but 503s.... I went through a lot of trial and error on this thing and well, apparently I left a "spiderweb" hanging :)

PR #931, opened 2022-08-26
Param val

bflynn-collabralink:
Story: None

### Details

A deploy to Prod revealed that a StringList param type will attempt to be loaded as a string array when multiple values including commas are used as the param value. This param specifically is then used to load as an environment variable into a lambda which cannot handle a string array. Therefore we need to use a basic String type and rely on the app to parse the values.

### Changes

- update CmsSpaFormCmsEmail to `Type: String`
`
### Implementation Notes

- We will want in the future to change all other email based params (or any being used as env var in lambda) to also be String type.

### Test Plan

1. After deploy replace the default value with an email list that includes commas in the value
2. Submit a chip spa and ensure the email is delivered


bflynn-collabralink:
This is the equivalent change we already made in val and pushed to prod.

Dark-Knight-000 (approved):

PR #945, opened 2022-09-02
Oy2 20220

ZacharyRizer:
Story: https://qmacbis.atlassian.net/browse/OY2-20220

### Details

Updating the tables to better display when there is information that overflows the boundaries of the table. 

### Changes

- Changed the tables to have a fixed height
- Doing this allows the bottom scroll bar to always be visible to better indicate when there is content off the x-axis of the table

### Implementation Notes

- Changed the css for table-wrapper in index.scss: doing this updates all the tables simultaneously


bflynn-collabralink (approved):
Looks good! Thanks for explaining why we moved max-width up as well!

PR #946, opened 2022-09-06
OY2-20220 test fixes

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-20220

### Details
Fixing tests that were failing due to the scroll. I tried a lot of things, but sometimes in the automation, the page is autoscrolling to the bottom and the program can't scroll back up to see or click on an item. I ended up having to do force clicks which isn't ideal at all, but was all I could see as a solution since I wasn't able to reproduce the issue while manually testing.


kristin-at-theta (approved):
I'm not going to pretend to understand the nuances, but it makes sense that scrolling is now important.  

Changes are confined to tests and they ran green on the branch!  Great work, Valencia :) 

Approved

PR #950, opened 2022-09-07
OY2 20383: Use EUA jobcodes, not IDM roles, to identify CMS read-only users

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-20383

### Details
OneMAC assumed EUA users would not have "cms-roles" from the Authenticating service (that used to be specific to IDM).  Now that field is present in some EUA users, so we have to change our strategy.

### Changes
- added ismemberof to the cognito user profiles (this holds the jobcodes)
- also updated our system to strip out cms-roles that are not oneMAC related.
- Took out the github contributers from the test users and hard-coded the dev users... I'm starting to get nervous that there's a lot of people who are no longer on the project on that list....
- hardcoded dev users have various cms-role values for testing

### Test Plan
1. Delete items for k.grue.cmsroleapprover@gmail.com, helpdesk@notreal.com, k.grue.stateuser@gmail.com, helpdesk2@notreal.com, k.grue.stateadmn@gmail.com
2. Login as "k.grue.cmsroleapprover@gmail.com"
3. Should end up a CMS Read-only User with no errors
5. Login as "helpdesk@notreal.com"
6. There will probably be an error that flashes past, but application should keep at it until a successful pending HelpDesk role is reached.
7. k.grue.stateuser@gmail.com
8. Should be sent directly to the state signup choices
9. helpdesk2@notreal.com
10. There will probably be an error that flashes past, but application should keep at it until a successful pending HelpDesk role is reached.
11. k.grue.stateadmn@gmail.com
12. Should be sent directly to the state signup choices


bflynn-collabralink (approved):
Looks ok as far as I can tell. Lets see what happens!

PR #969, opened 2022-09-15
Dependabot: Update axios and amplify

kristin-at-theta:
Story: Dependabot

### Details
amplify needed to be upgraded.... this was a breaking change, so had to fix the breaks!

### Changes
- Updated amplify package version in package.json
- fixed resulting API call errors in UserDataApi
- added an await to the logout to be better about logging out...

### Test Plan
1. Application should run as before....


bflynn-collabralink (approved):
Looks good to me!
Tested with developer login and everything behaved as expected.
Could not test with EUA/IDM user in a feature branch.
Approved!

PR #975, opened 2022-09-19
Oy2 20375

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-20375

### Details

Require new temporary extensions from New Submission button in package view to enter the approved initial or renewal waiver.

### Changes

- added parent validation lambda for temp waiver
- added config to temp extension types to use existing parent id entry infrastructure
- updated componentId to allow passing in an id so that parent component id can differ from the actual component id on the form when using more than one `ComponentId` component on a page

### Test Plan

1. Login as state submitter
2. Navigate to package view and New Submission -> Waiver Action -> Request Temporary Extension
3. Verify title and subtext of parent id field per ACs
4. Enter any parent id that is not an approved initial or renewal waiver
5. Verify the error message matches the AC


services/common/type/waiverTemporaryExtension.js
```diff
@@ -1,24 +1,19 @@
...
-    {
-      idMustExist: true,
-      errorLevel: "warn",
-      validateParentId: true,
-    },
   ],
   allowMultiplesWithSameId: false,
   requiredAttachments: ["Waiver Extension Request"],
   optionalAttachments: ["Other"],
   requiredUploads: ["Waiver Extension Request"],
```
kristin-at-theta, line 15: can get rid of the uploads vars now, I think :)

services/ui-src/src/components/ComponentId.tsx
```diff
@@ -67,7 +69,7 @@ const ComponentId: React.FC<{
           <input
             className="field"
             type="text"
-            id="componentId"
+            id={id}
```
kristin-at-theta, line 72: Nice value-add :)

services/ui-src/src/page/OneMACForm.tsx
```diff
@@ -165,10 +165,10 @@ const OneMACForm: React.FC<{ formConfig: OneMACFormConfig }> = ({
     updatedRecord.territory = getTerritoryFromComponentId(componentId);
 
     //if this is a child type form and no parentId was passed in state then determine parentId based on componentId
-    if (typeof formConfig.getParentInfo == "function" && !presetParentId) {
-      [updatedRecord.parentId, updatedRecord.parentType] =
-        formConfig.getParentInfo(updatedRecord.componentId);
-    }
+    // if (typeof formConfig.getParentInfo == "function" && !presetParentId) {
```
kristin-at-theta, line 168: if we are truly done with getParentInfo, should remove the code :)

services/ui-src/src/page/OneMACForm.tsx
```diff
@@ -9,7 +9,7 @@ import React, {
 import { useHistory, useLocation } from "react-router-dom";
 import { Input } from "rsuite";
 
-import { TextField, Button, Dropdown, Review } from "@cmsgov/design-system";
+import { TextField, Button, Dropdown } from "@cmsgov/design-system";
```
kristin-at-theta, line 12: this confused me at first, but I see that the Review piece is also in ComponentId, so nice cleanup!

kristin-at-theta (approved):
Nicely done and I like that you cleaned up those duplicate review elements.  I also like that you fixed the "duplicate id" bug, though now we have one less story next sprint....:) 

Approved, though I wouldn't mind it if you were to delete out some of those commented sections..... :)

PR #977, opened 2022-09-20
Dependabot: Update react-scripts

kristin-at-theta:
Story: Dependabot

### Details

### Changes
- updated react-scripts to 5.0.1

### Test Plan
1. Build should deploy
2. Application should run as before


.github/workflows/deploy.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         run: npx eslint --max-warnings=0 services
   unit-test:
     name: Unit Test
-    runs-on: macos-latest
```
bflynn-collabralink: 😮

bflynn-collabralink (approved):
oh boy thats a lot of test updates......im sorry :(

app seems to work as before...approved!

.github/workflows/deploy.yml
kristin-at-theta: That got changed back!  It was part of the experimenting :). apparently the macos is more stable and 10 times more expensive (which I didn't find out until after I tried it and yes, it worked :) )

PR #981, opened 2022-09-22
Oy2 20404 - unique component id

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-20404

### Details

Add custom id attributes to componentId component when using multiple of same component on a page

### Changes

- Added optional `idPrefix` param to componentId
- updated componentId to use idPrefix if present for all ids
- updated OneMacForm.tsx to use idPrefix for parent component id

### Test Plan

1. Login as state submitter
2. Navigate to New Submission -> Waiver Action -> Renewal, Amendment, or Temp Extension
3. Verify the html id for the parent component id fields and the main component id fields are unique
4. Ensure id uniqueness for labels, fields, grey subtext, and any error text

### DIFF
--- a/services/ui-src/src/components/ComponentId.tsx
+++ b/services/ui-src/src/components/ComponentId.tsx
@@ -73,7 +73,7 @@ const ComponentId: React.FC<{
             name="componentId"
             aria-describedby={idFieldHint
               ?.map(function (idFieldHint, idx) {
-                return "fieldHint" + idx;
+                return idPrefix + "fieldHint" + idx;
               })
               .join(" ")}
             value={value}
--- a/services/ui-src/src/components/ComponentId.tsx
+++ b/services/ui-src/src/components/ComponentId.tsx
@@ -8,6 +8,7 @@ import { Message } from "../libs/formLib";
  * Returns the ID specific form element
  */
 const ComponentId: React.FC<{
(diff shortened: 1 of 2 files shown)


bflynn-collabralink:
Note: 2 of the failing Cypress tests are expected due to changes in the id values

services/ui-src/src/components/ComponentId.tsx
```diff
@@ -69,7 +69,7 @@ const ComponentId: React.FC<{
           <input
             className="field"
             type="text"
-            id={id}
+            id={idPrefix + "componentId"}
             name="componentId"
             aria-describedby={idFieldHint
```
kristin-at-theta, line 74: Should add the idPrefix to the described-by field as well - currently the form has both of them with fieldhint0...

kristin-at-theta (changes requested):
Looks great!  I checked the hints, errors, etc.

I do think the aria-describedby should also receive the prefix, so am requesting that change :)

services/ui-src/src/components/ComponentId.tsx
bflynn-collabralink: fixed!

kristin-at-theta (approved):

PR #982, opened 2022-09-22
OY2-20691 - update detail card format

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-20691

### Details

Update Package Details page to add Status label above the current package status

### Changes

- add Review item surrounding status
- created css styles for status and for top level review items in a card (in order to manage whitespace)
- make package actions also a Review item to get consistent header labels

### Test Plan

1. Login as a state submitter
2. Navigate to Package List
3. Click any item in the list to view the Details page
4. Verify status label and package actions label look as specified


services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -67,25 +70,26 @@ export const DetailSection = ({
...
-              <h2>{pageConfig.actionLabel}</h2>
-              <ul className="action-list">
-                {pageConfig.actionsByStatus[detail.currentStatus]?.length >
-                0 ? (
-                  pageConfig.actionsByStatus[detail.currentStatus]?.map(
-                    (actionName, i) => (
-                      <li key={i}>
-                        {actionComponent[actionName](detail, updateData)}
-                      </li>
+              <Review heading={pageConfig.actionLabel}>
```
kristin-at-theta, line 73: I feel like it makes sense to have the Status label be a Review heading and I know the styles are the same for the Package Actions label.... but.... I am concerned that "Package Actions" is more of a section header than an item header for screen reading purposes?  Anyway, I noticed it and thought it was worth a mention....

kristin-at-theta (approved):
Looks good - see my comment about the screen readers and sections v labels, but I don't think that's going to be a problem...

Approved!

PR #986, opened 2022-09-27
OY2 20703: Implement New Serverless Waf Plugin

kristin-at-theta:
Story: [Dependabot](https://qmacbis.atlassian.net/browse/OY2-20703

### Details
QuickStart uses a plugin to manage the wafs, use that plugin in the OneMAC repo.

### Changes
- added the plugins
- updated workflows to use yarn instead of npm to more closely match QuickStart and hopefully remove some of our package dependency issues/errors.
- updated the node version in the workflow to look for the .nvmrc file
- updated the caching of dependencies to match QuickStart

### Test Plan
1. Deploy should be mostly green
2. Application should work as before


bflynn-collabralink:
Just a random observation when comparing the cms waf rules to our previous custom rules.......the cms waf does not include American Somoa in the list of allowed countries?!?!

bflynn-collabralink (approved):
Wow this is a lot. I like how you brought in using the nvmrc config so that its only one property in one place! I dont quite understand why yarn solved our woes but good job figuring that out!

I truly dont know how to test things other than "it works"....but see my comment about the waf countries. Thats all i got. :)

Approved!

PR #997, opened 2022-09-30
Oy2 20967 - temporary extension form updates

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-20967

### Details

Update lables and add temporary extension type form field

### Changes

- Updated form title
- updated id label and faq label
- added temporary extension type field
- - added new property to OneMacFormConifg and OneMacFormData
- - created component to wrap implementation and added component to form
- - added to joi schema for submitWaiverExtension

### Test Plan

1. Login as state submitter
2. Navigate to package view and click New Submission button
3. Select Waiver Action -> Request Temporary Extension
4. Verify on form the title and labels match ACs
5. Verify the new field "Temporary Extenstion Type" has the appropriate options and is required
6. Verify the form will not submit without selecting a Temporary Extension Type
7. Verify form will submit after entering all required fields (including the Temporary Extension Type)

### DIFF
--- a/services/app-api/email/CMSSubmissionNotice.test.js
+++ b/services/app-api/email/CMSSubmissionNotice.test.js
@@ -19,5 +19,5 @@ it("builds the CMS Submission Notice Email", async () => {
 
   expect(response2.HTML.includes(testData.componentId)).toBe(true);
   expect(response2.HTML.includes(process.env.applicationEndpoint)).toBe(true);
-  expect(response2.HTML.length).toBe(1222);
+  expect(response2.HTML.length).toBe(1231);
 });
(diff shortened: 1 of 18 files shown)


kristin-at-theta (changes requested):
Front end look great - looks like the Typescript is getting more organized.... I'll have to learn that, eh?

the required stuff and form submitting works great, but I don't see the new attribute in the database and if I select 1915(c), the details always comes back with 1915(b) (guessing it is the default)

I don't think the attribute is getting to the email, either :)

bflynn-collabralink:
> https://dhi567bpgtpfj.cloudfront.net

Good catch on the emails, ummmm it isnt in the ACs 🤣 
Added to the emails and adjusted tests.

Also fixed the final piece for it to save completely in the DB......not sure how i let that happen 😬

kristin-at-theta (approved):

PR #998, opened 2022-10-03
OY2 20983: Package Details Page Timestamp - All Details Pages

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-20983

### Details
Date Submitted on the packages needs to have the time as well as the date shown.

### Changes
- Updated Initial Submission Date format to include the time
- updated the RAI header to also include the time, since it is a Submission Date
- left the "date picker" dates without times, since we don't have those

### Test Plan
1. Log in as a submitting user
3. Find a package of each type, verify that the package details page uses the correct date/time format

### DIFF
--- a/services/ui-src/package-lock.json
+++ b/services/ui-src/package-lock.json
@@ -23,6 +23,7 @@
         "cmscommonlib": "file:../common",
         "core-js": "^3.7.0",
         "date-fns": "^2.16.1",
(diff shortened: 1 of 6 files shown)


bflynn-collabralink (approved):
Looks good to me!

Good thinking on the TE submission date titles from the parent details page....I never would have thought of that!

Approved!

PR #1000, opened 2022-10-04
OY2-21003 - change withdraw to withdraw package

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-21003

### Details

Update actions from "Withdraw" to "Withdraw Package" and also update confirmation dialog with the same.

### Changes

- Updated Withdraw component to show desired language
- Also updated popup component since TE still uses that from the mini dash (couldnt consolidate easily)
- Updated tests to reflect language changes

### Implementation Notes

- I tried to make TE use the same ActionPopup component that Package list uses however it messed too much with the tests since I dont believe we still found a way to hit buttons in the popup since it is outside the context of the test. Furthermore it is my understanding the mini dash will be going away in favor of moving all items to top level items on the package list. So its a moot point anyway and we can retire the older popup component at that time.

### Test Plan

1. Login as statesubmitter
2. navigate to package dashboard and verify actions for types: Initial Waiver, Renewal, Amendment, and Temporary Extensions that are in a withdrawable status (such as submitted or in review) have the action listed as "Withdraw Package"
3. navigate to the details page and verify the action is also "Withdraw Package"
4. click the withdraw package link on the details page (could also be from the kebab on the package list) and verify the popup is named Withdraw Package.

### DIFF
--- a/services/ui-src/src/components/PopupMenu.js
+++ b/services/ui-src/src/components/PopupMenu.js
@@ -28,8 +28,8 @@ const VARIATION_PROPS = {
     width: "wide",
   },
   PackageList: {
-    acceptText: "Yes, withdraw",
-    dialogTitle: "Withdraw?",
+    acceptText: "Yes, withdraw package",
+    dialogTitle: "Withdraw Package?",
   },
 };
 
--- a/services/common/index.d.ts
+++ b/services/common/index.d.ts
@@ -102,7 +102,7 @@ export namespace ChangeRequest {
 
   export enum PACKAGE_ACTION {
     RESPOND_TO_RAI = "Respond to RAI",
-    WITHDRAW = "Withdraw",
+    WITHDRAW = "Withdraw Package",
     REQUEST_TEMPORARY_EXTENSION = "Request a Temporary Extension",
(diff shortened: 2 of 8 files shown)


services/ui-src/src/components/PopupMenu.js
```diff
@@ -28,8 +28,8 @@ const VARIATION_PROPS = {
     width: "wide",
   },
   PackageList: {
-    acceptText: "Yes, withdraw",
-    dialogTitle: "Withdraw?",
+    acceptText: "Yes, withdraw package",
```
kristin-at-theta, line 31: I agree with the capitalization choice.... but it looks kinda odd !

kristin-at-theta (approved):
Looks great!  Yes... some of that confirmation dialog thing was all because of the test coverage and well, the REAL solution is to change the confirmation dialog to return the answer to the controlling component... BUT... that's actually hard (for me) to accomplish when you can also assign a function pointer and it does it for you (but you can't test it)

Anywhoo, Approved :)   I checked that all the types have the Withdraw Package text and the popup looks correct.  I did not withdraw anything :P

PR #1001, opened 2022-10-04
OY2 20217: Consolidate a list of all messages and correlating forms for Submission and Package View

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-20217

### Details
A list of the messages and form messages, etc will be obsolete pretty much the moment it is created.... devise an automagic form describer instead - at least for the Package View, which can change more.

### Changes
- cleaned up SOME of the error messaging in the forms 
- created "forms-describe" - a hidden page which describes package view forms
- removed all the complicated ID existence checking carried over from Submission View - Package View will have a better concept of "parentage" so should not need the ability to perform extra checks on ID existence based on format alone.

### Implementation Notes
- it would be super great if this did not exist in the application, if it could be some sort of admin tool that references the config files and supporting messaging, instead...

### Test Plan
1. Login to the Application as a System Admin User
2. Navigate to forms-describe using the URL address
3. see all the goodies!

### DIFF
--- a/services/app-api/form/submitAny.js
+++ b/services/app-api/form/submitAny.js
@@ -35,11 +35,6 @@ export const submitAny = async (event, config) => {
     throw error;
   }
 
-  // const data = {
-  //   ...inData,
-  //   submitterId: event.requestContext.identity.cognitoIdentityId, // not sure we want this anymore
-  // };
-
   // errors here are application level: returned as codes to front end for handling
   try {
     // returns undefined if no errors found, or the first error found.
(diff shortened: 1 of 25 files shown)


services/app-api/form/submitAny.js
```diff
@@ -68,42 +63,15 @@ export const submitAny = async (event, config) => {
...
-      .filter((errorMsg) => errorMsg != undefined)
-      .forEach((errorMsg) => {
-        console.log("theLevel: ", errorMsg[0]);
-        console.log("theCode: ", errorMsg[1]);
-        if (errorMsg[0] === "error") {
-          throw errorMsg[1];
-        } else {
-          warningsInCMSNotice.push(errorMsg[1]);
-        }
-      });
```
bflynn-collabralink, line 106: I love this cleanup! At first i was thrown by warningsInCmsNotice being dropped......but since we no longer have warnings in package view it makes sense.

services/common/type/chipSPA.js
```diff
@@ -5,12 +5,7 @@ export const chipSPA = {
   idLabel: "SPA ID",
   idRegex:
     "(^[A-Z]{2}-[0-9]{2}-[0-9]{4}-[a-zA-Z0-9]{1,4}$)|(^[A-Z]{2}-[0-9]{2}-[0-9]{4}$)",
-  idExistValidations: [
-    {
-      idMustExist: false,
-      errorLevel: "error",
-    },
-  ],
+  idMustExist: false,
```
bflynn-collabralink, line 8: I feel like you did such a good job in reducing idExistValidations into a single boolean that I also wonder if now idMustExist and allowMultiplesWithSameId mean the same thing?? Just a thought for another time...

services/ui-src/src/page/OneMACForm.tsx
```diff
@@ -229,98 +217,29 @@ const OneMACForm: React.FC<{ formConfig: OneMACFormConfig }> = ({
...
+      );
+      if (validationMessages.length === 0 && oneMacFormData.componentId) {
+        try {
+          const isADup = await ChangeRequestDataApi.packageExists(
+            oneMacFormData.componentId
+          );
+          if (isADup === false && formConfig.idMustExist) {
+            validationMessages.push(buildMustExistMessage(formConfig));
+          } else if (isADup === true && !formConfig.idMustExist) {
+            validationMessages.push(buildMustNotExistMessage(formConfig));
```
bflynn-collabralink, line 232: amazing job condensing this!!

services/ui-src/src/page/DescribeForms.tsx
```diff
@@ -0,0 +1,589 @@
+import React from "react";
```
bflynn-collabralink, line 1: This whole component and its subcomponents are just brilliant......there is no way I could have done it this well. Bravo!

bflynn-collabralink (approved):
WOW.....this is super amazing. Im not sure what else to say. I didnt spot any logic flaws and I love the consolidation you did with the id validation. Then I got to the DescribeForms component and was amazed. Great job!
APPROVED!!

bflynn-collabralink:
Oh forgot to mention you might want to update the test instructions to note that you have to be logged in specifically as a cms system admin. I dug through the code to figure it out but I did wonder why it wouldnt load the page at first.

services/common/type/chipSPA.js
kristin-at-theta: So... I agree.... and I think they CAN be combined at this time.... BUT - I feel like we will eventually need to handle that Medicaid SPAs can only have one RAI/Response to RAI while CHIPs can have multiple.... though, that should be handled by CMS, but.... I guess I chickened out on merging that up.... :)

kristin-at-theta:
Oh, ha ha - yes.... I added the extra check on the role *after* I made the draft PR and forgot to update!  Thanks for the catch and for the thorough review :)

PR #1004, opened 2022-10-07
Oy2 20996 - remove 90th day

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-20996

### Details

Remove 90th day from details page
Remove 90th day and expiration date from package list

### Changes

- Removed column definitions from package list
- configured default detail config to `show90thDayInfo: false,` (if you prefer we just yank everything out altogether that is easy as well)

### Test Plan

1. Login as a statesubmitter
2. Navigate to package view and verify columns are no longer present in the grid/filter/show hide options
3. Navigate to details pages for a package in every status (since that is what used to drive the display) and verify 90th day no longer exists in the card

### DIFF
--- a/services/ui-src/src/containers/PackageList.test.js
+++ b/services/ui-src/src/containers/PackageList.test.js
@@ -3,13 +3,12 @@ import React from "react";
 import {
   render,
   screen,
-  within,
   waitForElementToBeRemoved,
 } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { MemoryRouter } from "react-router-dom";
 
-import { ONEMAC_ROUTES, Workflow } from "cmscommonlib";
+import { ONEMAC_ROUTES } from "cmscommonlib";
 import { AppContext } from "../libs/contextLib";
 import {
   stateSubmitterInitialAuthState,
@@ -104,7 +103,6 @@ it("renders table with spa columns", async () => {
   screen.getByText("SPA ID");
   screen.getByText("Type");
   screen.getByText("State");
-  screen.getByText("90th Day");
   screen.getByText("Date Submitted");
   screen.getByText("Submitted By");
   screen.getByText("Actions");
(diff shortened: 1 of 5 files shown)


services/ui-src/src/index.scss
```diff
@@ -1005,6 +1005,7 @@ article.form-container {
 
   @media only screen and (min-width: 1024px) {
     display: flex;
+    flex-wrap: wrap;
```
kristin-at-theta, line 1008: Sneaking in that "No Results" wrapping fix, I see :) nice value add!

services/ui-src/src/libs/detailLib.ts
```diff
@@ -101,7 +101,7 @@ export const defaultDetail: OneMACDetail = {
     },
   ],
   actionsByStatus: Workflow.defaultActionsByStatus,
-  show90thDayInfo: true,
+  show90thDayInfo: false,
```
kristin-at-theta, line 104: Based on the hmmm-ing and haw-ing at refinement over if they would ever bring this back, I agree it is better to disable instead of removing.

services/ui-src/src/containers/PackageList.test.js
```diff
@@ -152,34 +150,34 @@ it("does not display waiver rais", async () => {
...
-    await waitForElementToBeRemoved(() => screen.getByTestId(LOADER_TEST_ID));
-
-    // get the row for the status
-    const packageRow = within(
-      screen.getAllByText(filterFieldValue)[0].closest("tr")
-    );
-    expect(packageRow.getAllByText(textShown)[0]).toBeInTheDocument();
-  }
-);
+// it.each`
```
kristin-at-theta, line 153: Part of me wants to delete this, but.... it is actually a pretty good example of doing this kind of compound test with different data and I'm pretty sure I've copied it before :)

kristin-at-theta (approved):
Nice and clean, great choices in what to disable and what to remove!  Nice fix of the table wrapping, too!

Approved!

services/ui-src/src/containers/PackageList.test.js
bflynn-collabralink: yeah i really liked it as an example too and was going to make a comment about why i left it in...but forgot. :)

PR #1006, opened 2022-10-07
OY2 21051: [Package Dashboard] Error sending submission confirmation email

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21051

### Details
The state submitter receipt emails were not getting sent from Package form submissions.  The error in AWS suggested that SES was expecting an array of To Addresses and we sent a string.

### Changes
- converted the ToAddress field to an array
- tweaked the Component ID styles because Temporary Extension Number is long and SPA ID is short and they need to both look appropriate.

### Test Plan
1.Login with a submitting user who has a real email that you can access. (I have some gmail accounts expressly for this purpose, let me know if you have one you would like added to the Test Users list)
2. Navigate to Package View
3. Submit a form
4. Verify you receive the state submitter receipt for your form submission.

### DIFF
--- a/services/app-api/email/stateSubmissionReceipt.js
+++ b/services/app-api/email/stateSubmissionReceipt.js
@@ -14,7 +14,7 @@ export const stateSubmissionReceipt = (data, config) => {
   data.attachments = {};
 
   return {
-    ToAddresses: `${data.submitterName} <${data.submitterEmail}>`,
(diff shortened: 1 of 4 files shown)


bflynn-collabralink (approved):
Looks good! Nice work on the adjustment of the label layouts for the ComponentId!!

PR #1007, opened 2022-10-11
DevOps: use file .nvmrc to control node version

kristin-at-theta:
Story: None, is in quick-start

### Details
There are several quick-start updates that are not reflected in the OneMAC repo - we tried a really large update and had to roll back, so am now doing these in much smaller bites.

### Changes
- Created the file .nvmrc to hold the node version for use in the workflows
- updated the checkout to v3
- updated the node version to use the file

### Test Plan
1. Repo should deploy
2. Automated tests should be green
3. Application should run as before.


kristin-at-theta:
deleting this branch.... in order for .nvmrc to work, it uses the deprecated action set-output... suggestion from github is to move to environment files, instead.

PR #1012, opened 2022-10-18
replace survey success banner with normal success banner

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-21002

### Details

Replace survey success banner with normal success banner

### Changes

- all success banner messages were configured in one spot in alertlib for both submission and package view

### Test Plan

1. Login as a statesubmitter
2. submit a new package from both submission view and package view
3. verify success banner after submission does not contain the survey language or link


kristin-at-theta (approved):
Yup!  That's what I would've done!

Took the extra time to test it in feature branch (at least for Medicaid SPA form, both dashboards) and it works :) 

Approved!

PR #1022, opened 2022-10-28
OY2 21350: Issues When exporting to the User Management Screen

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21350

### Details
when the dates were switched from seconds to milliseconds, the seed data and the export to CSV were missed.

### Changes
- removed the * 1000 in the export, since data is now stored in milliseconds
- updated seed data to use milliseconds for user date (basically a changeDate)

### Test Plan
1. Login as a user with user management abilities
2. Navigate to User Management Dashboard
3. Verify the Dates shown
4. Click "Export to CSV"
5. Open the CSV
6. Verify the Dates


bflynn-collabralink (approved):
looks. good to me! ....I dont know if Padma will remember or be the one to do QE on this but might want to add the comment to the story that CMS Read Only User is a valid role name since that is still in the ACs.
Approved!

PR #1023, opened 2022-10-31
OY2 21386: Package View- Need Formal RAI Response Section on Detail Page of 1915(c) Appendix K

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21386

### Details
Appendix K detail pages are not showing the RAI responses for the App K

### Changes
- Updated the waiver RAI type logic to not include waiverappk so waiverappkrai types would be found

### Test Plan
1. Login as a user that can view package details
2. Navigate to Package Dashboard -> Waivers tab
3. Find an App K in RAI Issued status (no way to tell from the dashboard if it already has an RAI Response)
4. Click into the App K details page
5. Verify that the RAI Responses are there in the same way that other RAI Responses are displayed
6. If there is no RAI Response displayed, use the Respond to RAI action to add one.


bflynn-collabralink (approved):
As you said, it works!

Approved.

PR #1029, opened 2022-11-03
OY2-21321 Test updates

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-21321

### Changes

- updated tests to include modal check
- also added in text validation for the modals in all RAI responses per Jaime's comment


kristin-at-theta (approved):
Looks good :) 

Approved!

PR #1030, opened 2022-11-04
OY2 21351: Package View - Email Blast to State Submitters/State Sys Admins when Package is Withdrawn

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21351

### Details
When a package is withdrawn, ALL active submitting users for that territory should receive the email notification.

### Changes
- Added a database query to find the active state submitters and state system admin emails
- reconfigured the email sending to be able to handle the asynchronous query
- moved some functionality into the config for readability/ease of finding business requirements
- Added use of the GSI2 index on "user role items" to be able to query the active users for state submitters (couldn't use the contactInfo way we use for cms approvers because state submitters can have different statuses for different territories all connected to one email address)
- Updated the seed data
- Created a migrate script to handle data changes moving up the environments.

### Implementation Notes
- BRIEFLY considered an aggregate item to hold the active email list, updated as users are added and/or their statuses changed ... this *would* be an excellent practice use case for using the stream to update aggregates, BUT... it's also a major paradigm shift that I'm not sure we want to do just yet....

### Test Plan
1. Log in as a Submitting User with a valid email address
3. Withdraw a package
4. Verify you got the Withdrawal Email Receipt.

### DIFF
--- a/services/app-api/action/changeStatusAny.js
+++ b/services/app-api/action/changeStatusAny.js
@@ -3,8 +3,6 @@ import { RESPONSE_CODE } from "cmscommonlib";
 import { getUser } from "../getUser";
 import { validateUserSubmitting } from "../utils/validateUser";
 import updateComponent from "../utils/updateComponent";
-import { CMSWithdrawalNotice } from "../email/CMSWithdrawalNotice";
-import { stateWithdrawalReceipt } from "../email/stateWithdrawalReceipt";
 import sendEmail from "../libs/email-lib";
 
 export const changeStatusAny = async (event, config) => {
@@ -43,11 +41,13 @@ export const changeStatusAny = async (event, config) => {
   }
 
   try {
-    await Promise.all(
-      [CMSWithdrawalNotice, stateWithdrawalReceipt]
-        .map((f) => f(updatedPackageData, config))
-        .map(sendEmail)
+    const theEmails = await Promise.all(
+      config.emailFunctions.map(
+        async (f) => await f(updatedPackageData, config)
+      )
     );
+    console.log("the Emails: ", theEmails);
+    theEmails.map(sendEmail);
   } catch (e) {
     console.error("Failed to send acknowledgement emails", e);
   }
--- a/services/app-api/action/defaultWithdrawConfig.js
+++ b/services/app-api/action/defaultWithdrawConfig.js
@@ -1,6 +1,10 @@
 import { Workflow, RESPONSE_CODE } from "cmscommonlib";
 
+import { CMSWithdrawalNotice } from "../email/CMSWithdrawalNotice";
+import { stateWithdrawalReceipt } from "../email/stateWithdrawalReceipt";
+
 export const defaultWithdrawConfig = {
   newStatus: Workflow.ONEMAC_STATUS.WITHDRAWN,
   successResponseCode: RESPONSE_CODE.PACKAGE_WITHDRAW_SUCCESS,
+  emailFunctions: [CMSWithdrawalNotice, stateWithdrawalReceipt],
 };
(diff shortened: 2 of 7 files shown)


services/app-api/email/stateWithdrawalReceipt.js
```diff
@@ -1,17 +1,69 @@
...
+    Subject: `${config.typeLabel} Package ${data.componentId} Withdraw Request`,
+    HTML: `
+        <p>This is confirmation that you have requested to withdraw the package below. The package will no longer be considered for CMS review:</p>
+        ${formatPackageDetails(data, config)}
+        <p>Thank you!</p>
+        `,
+  };
+  try {
+    stateReceipt.ToAddresses = await getAllActiveStateUserEmailAddresses(
+      data.componentId.substring(0, 2)
```
bflynn-collabralink, line 62: I "think" we should have id.territory here since we are doing the prior update using `ALL_NEW`? It shouldnt make a difference but i think it would be more clear in addition to working for any legacy data we may import that didnt follow correct format? If its not available then not worth it, and also just posing the question for consideration and not officially "requesting a change". :)

services/app-api/libs/email-lib.js
```diff
@@ -40,13 +40,10 @@ function getSESEmailParams(email) {
 export default function sendEmail(email) {
   try {
     const emailParams = getSESEmailParams(email);
-    // If we are in offline mode just log the email message.
+    console.log("Amazon email parameters: ", emailParams);
+    console.log("and the message is: ", emailParams.Message.Body.Html);
```
bflynn-collabralink, line 44: I really like that you moved this outside the local processing block!

services/app-api/action/changeStatusAny.js
```diff
@@ -43,11 +41,13 @@ export const changeStatusAny = async (event, config) => {
   }
 
   try {
-    await Promise.all(
-      [CMSWithdrawalNotice, stateWithdrawalReceipt]
-        .map((f) => f(updatedPackageData, config))
-        .map(sendEmail)
+    const theEmails = await Promise.all(
+      config.emailFunctions.map(
```
bflynn-collabralink, line 45: Did we define emailFunctions on other events that call changeStatusAny? I only see it on defaultWithdrawConfig?

services/app-api/action/defaultWithdrawConfig.js
```diff
@@ -1,6 +1,10 @@
 import { Workflow, RESPONSE_CODE } from "cmscommonlib";
 
+import { CMSWithdrawalNotice } from "../email/CMSWithdrawalNotice";
+import { stateWithdrawalReceipt } from "../email/stateWithdrawalReceipt";
+
 export const defaultWithdrawConfig = {
   newStatus: Workflow.ONEMAC_STATUS.WITHDRAWN,
   successResponseCode: RESPONSE_CODE.PACKAGE_WITHDRAW_SUCCESS,
+  emailFunctions: [CMSWithdrawalNotice, stateWithdrawalReceipt],
```
bflynn-collabralink, line 9: I love that this is in config now!

services/app-api/action/changeStatusAny.js
bflynn-collabralink: looks like we only use this for withdraws so far......so all is good!

bflynn-collabralink (approved):
Nice job on the new query! I like some of the optimizations you made such as always logging email contents and moving what emails will be sent into the config!

I did make one optional suggestion regarding usage of component id vs territory.

Approved!

services/app-api/email/stateWithdrawalReceipt.js
bflynn-collabralink: ahhh....for some reason i thought there was! Sounds good then!

PR #1031, opened 2022-11-06
Oy2 21405 - update error text for renewal id

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-21405

### Details

Waiver Renewal Number should include the text “the “R##” starts with ‘R01’ and ascends.” 

### Changes

- Update error text for Renewals Id

### Implementation Notes

- I updated based on existing language used on Amendments, so the text differs from that in the AC. I brought this up to the product analyst and product owner.

### Test Plan

1. Login as state submitter
2. Navigate to package view and click New Submission
3. Click Waiver Action -> 1915(b) Waiver Renewal
4. Enter a value for 1915(b) Waiver Renewal Number that begins with an appropriate state but has invalid format (Entering just the state portion alone would qualify)
5. Verify the error message now has the additional language regarding the R## starts with 01 and ascends.


kristin-at-theta (approved):
Yup!  Approved :)

PR #1047, opened 2022-11-22
OY2 21087: Data Migration for the unidirectional package View

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21087

### Details
This branch reorganizes the database to treat actions and submissions (from any source) as events and update the package entities accordingly.  Duplicate events are ignored, updated events update the entity (if the most current)

### Changes
- Submit functions now submit events
- new one-stream services watches the dynamodb stream and updates packages
- updating a package also updates it's parent, if it has one.
- added the retrieval of Parent Package Type to submitAny with a util of getPackageType
- admin: resetOneTable removes all items from one table, preserving user items
- admin: convertChangeRequests scans a changeRequest table and creates the events for them 
- admin verifyChangeRequests scans a changeRequest table and verifies that they each relate to a package with a GSI1pk of OneMAC#<tab>
-added the migration to the workflow for all but production (means the automated tests will use the migrated data)
- fixed a bug in the details section that was keeping additionalInformation from showing for amendments

### Test Plan
1. Branch should deploy
2. Use AWS console to view submission in the product-change-2 table and verify that they have been migrated into the appropriate format in the one-table


bflynn-collabralink (approved):
I think we talked initially about all my concerns and I havent found anything since. This is amazing work!

Approved.

PR #1059, opened 2022-12-07
OY2 21891: Update Latest Status from Sea Tool

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21891

### Details
Ingest SEATool events past and present and update the package accordingly

### Changes
- Updated buildPackage to include SEATool sourced events
- update resetOneTable so that it does NOT remove the SEATool events
- removed the ConditionExpression for the SEATool events, we do want them to update
- added seed data containing the "leftover" SEATool events

### Test Plan
1. Login to both the application and the AWS Console.
2. Use a GSI1pk search on "SEATool#Medicaid_SPA" (I also used a "Begins with" on the GSI1sk for "MD" )
3. Select a SEATool event and note the ID
4. do a pk query on that ID and see if any packages or OneMAC details exist.
5. If they do, verify that the package details in the application have the SEATool status
6. If not, try to submit with that number (and fail)
7. copy the dynamodb item and delete it
8. now submit with that ID
9. Create Item with the copied SEATool event (update the dates (CHANGED_DATE and STATUS_DATE and the sk timestamp) so they are later than the submission)
10. Verify that the package in the application has been updated to the new status.

Repeat for:
CHIP_SPA - same as Medicaid_SPA
1915b_waivers - with the waivers topic, you have to find the three action types, New, Renew, and Amend...
1915c_waivers topic should adjust app Ks, which are only the "Amend" action type


bflynn-collabralink (approved):
Looks good to me. I like the expansion of the parent validation and keeping that all within each types config!

Approved.

PR #1068, opened 2022-12-12
Oy2 22186 - remove left hand nav

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22186

### Details

Remove left hand navigation. Move additional details to bottom. Update TE submission landing page to package list.

### Changes

- removed all left hand nav functionality
- updated detailview to always include additional info section and removed conditional logic on detail section so that it never displays additional info section
- update TE form landing page

### Implementation Notes

- I thought about commenting it all out, but it was just too much for me to comfortably leave in. Source control will retain what was done if we ever put back a left hand nav.

### Test Plan

1. Login as any user with view privileges (such as state submitter)
2. Navigate to package details page for each package type and verify additonal information section always appears
3. When no additional information has been submitted then verify the default text stating as such.


kristin-at-theta (approved):
Wow.... a bit painful seeing all our good code just ripped out!  Though it still exists in history if we ever want it back.... I know the story says "hide," but I do agree with removing the code.  I feel like if/when they do want it back, it will be sufficiently different to mean using the old code won't be worth it.

In testing.... I find I do NOT like the Additional Information underneath the RAI Responses... BUT, that's what it states in the story AC *shrug*

Approved!

PR #1069, opened 2022-12-13
OY2 21899: Display Role based Status on package dashboard and details pages

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21899
Endpoint: https://d2b7wxo785vsd2.cloudfront.net

### Details
State users and CMS users see different package statuses

### Changes
- Migrated the sample data from product-change-request-2 into the feature branch change-request table
- adjusted the json files to point to change-request table (the default)
- Updated the Withdraw Action to update status to "Withdrawal Requested"
- Added flag on the role config for "isCMSUser" to designate which status mapping to use
- added lib/status-lib to hold the status mapping details
- Updated getDetail to convert the currentStatus to the status for the role
- getMyPackages now converts the currentStatus to the status for the role
- adjusted RAI Response submissions to update status to "Submitted"
- made the ONE_MAC statuses more "generic" and added in the new CMS statuses to be displayed
- removed the initial status list from Package List

### Test Plan
1. Log in as a state submitter
2. Verify the status labels in the package dashboard are the state user values
3. Verify the status labels in the package details are the state user values
4. Verify that Submitting an RAI returns the package to Submitted status
5. Verify that Withdrawing a package puts the package into Withdrawal Requested status
6. Log out and log in as a cms user (or, I used a different browser so I could view the same package with the two roles)
7. Verify the status labels on the package dashboard are the cms user values
8. Verify the status labels in the package details are the cms user values


services/common/workflow.js
```diff
@@ -26,16 +26,16 @@ export const ONEMAC_LABEL = {
...
-  RAI_SUBMITTED: "RAI Submitted",
-  IN_REVIEW: "Under Review",
-  RAI_ISSUED: "RAI Issued",
+  PENDING: "Pending",
+  PENDING_CONCURRANCE: "Pending - Concurrance",
+  PENDING_APPROVAL: "Pending - Approval",
+  RAI_ISSUED: "Pending - RAI",
   APPROVED: "Approved",
   DISAPPROVED: "Disapproved",
-  WITHDRAWN: "Package Withdrawn",
```
bflynn-collabralink, line 36: I suppose this doesnt matter since we are doing the mapping by user to completely different things and never showing these values, but just wondering why we dropped the Package off Package Withdrawn here?

bflynn-collabralink (approved):
Looks good and testing checked out as far as i could tell!

Approved!

services/app-api/action/defaultWithdrawConfig.js
```diff
@@ -4,7 +4,7 @@ import { CMSWithdrawalNotice } from "../email/CMSWithdrawalNotice";
 import { stateWithdrawalReceipt } from "../email/stateWithdrawalReceipt";
 
 export const defaultWithdrawConfig = {
-  newStatus: Workflow.ONEMAC_STATUS.WITHDRAWN,
+  newStatus: Workflow.ONEMAC_STATUS.WITHDRAWAL_REQUESTED,
```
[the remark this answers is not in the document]
kristin-at-theta: yeah.... in a way I was looking ahead here - I got my head all messed up around the idea of the same status being set by both OneMAC side and SEATool side and it probably wouldn't matter, but it was a relief when they agreed to the new status... (plus, this way, the story that "actually does" the withdraw updates will be easier to peer review and that will be during your vacation! :) )

services/common/workflow.js
kristin-at-theta: So... now that we have four sets of statuses and the true OneMAC status is only for viewing in the database (and for the mappings) ... I kinda felt being closer to the SEATool statuses would be more useful?

PR #1072, opened 2022-12-14
oy2-22186 test fixes

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-22186

### Details

Fix tests related and also including some general fixes since I am able to see most of the "real" test errors in Brandon's branch.


tests/cypress/cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "https://d2dr7dgo9g0124.cloudfront.net",
+  "baseUrl": "https://d2e4impk96qz2a.cloudfront.net",
```
kristin-at-theta, line 2: Probably want to change this back? or is it only used for local testing?

kristin-at-theta (approved):
Looks good!  Approved - though see my comment about the cypress.json file :)

tests/cypress/cypress.json
Valencia2019: It's local but I should change it back!

PR #1075, opened 2022-12-16
OY2 22238: Package actions to reflect approved actions matrix

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22238
Endpoint: https://d1n0hxr031pba5.cloudfront.net

### Details
following the spreadsheet in the jira ticket, verify that the Package Actions for each type, and status match the spreadsheet.

### Changes
- changed the OneMAC status text for some actions to match what the actions are looking for
- modified the "Actions matrix" to match the spreadsheet requirements
- updated the unit tests to match the new actions

### Implementation Notes
- Because only the state submitter has the ability to make actions at this time, we are keeping the action matrix in common and it is used by the front end... 
- In future, we should consider including the actions matrix in the response from the backend, as the backend is aware of the status choices for a particular role and we've removed that knowledge from the UI.

### Test Plan
1. Open the spreadsheet of actions
2. login as a submitting user
3. Navigate to package dashboard
4. Verify that packages in the statuses have the actions in the spreadsheet
5. also verify the package details pages for the correct actions.


bflynn-collabralink (approved):
Everything looks good as far as I can see.

Wanted to say thanks for manually putting in the endpoint to the PR!

Approved!

PR #1099, opened 2023-01-04
OY2 21948: Establish Framework for syncing data to OneMAC for SEA Tool entries in Pending RAI status

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-21948

### Details
We want to grant state users the ability to  submit RAI Responses on non-OneMAC packages in Pending-RAI status (SEA Tool).  To do this, we are adding a new "create" event to OneMAC, which is a type of submission event, one with a past submission date, no attachments, and no submitter information.

### Changes
- added the createOneMacPackage lambda to the admin service for adding a single new package.
- created the batchCreateOMP lambda in the admin service to allow a csv list of packages
- Updated the details section to display a package with no attachments

### Test Plan
1. Login to the AWS console
2. We assume these packages are already in SEA Tool, so best to find a SEA Tool package that is not (yet) in OneMAC.
3. Navigate to the createOneMacPackage lambda interface.
4. Use the sample JSON for the event, updating as needed to reflect the SEA Tool package.
5. run the Test event
6. Verify that the package has been created with the proper defaults and the SEA Tool Status (does not need to be Pending-RAI, but that would be bonus!)
7. Find a few more packages exclusively in SEA Tool
8. Create a csv "file" with their package details
9. navigate to the batchCreateOMP lambda interface
10. Use the sample JSON, replacing the csv value with the "file" contents
11. run the Test event
12. Verify that all packages in the csv were properly updated.

[NOSO.csv](https://github.com/CMSgov/onemac/files/10376319/NOSO.csv)


services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -110,12 +110,28 @@ export const DetailSection = ({
...
+                headingLevel="2"
+              >
+                <i>
+                  At this time, the attachments for this package are unavailable
+                  in this system. Contact your CPOC to verify the initial
+                  submission documents.
+                </i>
+              </Review>
+            </>
+          )}
```
bflynn-collabralink, line 134: This works. It would be nice though if the missing file message were integrated into the FileList component...maybe as a passed in customized message. Then we wouldnt have to duplicate the h2 header either.

bflynn-collabralink (approved):
This looks great! I love the use of a single run and batch run method!
I didnt verify any specific package as I took this story to be the mechanism to get any noso package (or group thereof) into onemac.

PR #1100, opened 2023-01-05
OY2 22310: Need to pull proposed effective date from SEA Tool

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22310

### Details
Submission View Submissions do not have Proposed Effective Date in their data.  SEA Tool does capture this date (though it is not a required field).  Because SEA Tool is the source of truth... the value in SEA Tool for PROPOSED_DATE is pulled into the package.  

NOTE: This may mean that a OneMAC package that was submitted with a Proposed Effective Date will "lose" that date if it is not properly entered into SEA Tool.

### Changes
- Update the Package details proposedEffectiveDate with the most recent SEA Tool value
- Use "none" if there is no PROPOSED_DATE included in the SEA Tool event
- Fixed the currentStatus disappearing while we wait for second SEA Tool event
- added a schedule to the convertChangeRequests lambda to manage migrations (testing and PROD)

### Test Plan
3. Login as a submitting user
4. Use the Submission Dashboard to create a new package
5. Go to AWS Console
6. Duplicate a SEATool event, update the ID_NUMBER to match the new package, update the timestamp in the sk to be a few milliseconds after the new package submission timestamp, and save the new event
7. Wait for convert lambda to run (every 15 minutes)
8. Verify that the package got created in OneMAC
9. Verify that the proposed effective date matches what is in SEA Tool ("none" if SEA Tool is null)
10. Feel free to use console to update the PROPOSED_DATE on the SEA Tool event and verify that the date on the package details updates to the new date.  (PROPOSED_DATE is not a required field in SEA Tool, so, in DEV, there are very few filled in)

### DIFF
--- a/services/one-stream/package/buildAnyPackage.js
+++ b/services/one-stream/package/buildAnyPackage.js
@@ -1,5 +1,7 @@
 const _ = require("lodash");
 import AWS from "aws-sdk";
+import { DateTime } from "luxon";
+
 import { dynamoConfig, Workflow } from "cmscommonlib";
 
 const dynamoDb = new AWS.DynamoDB.DocumentClient(dynamoConfig);
@@ -141,6 +143,9 @@ export const buildAnyPackage = async (packageId, config) => {
           seaToolStatus
         );
         putParams.Item.currentStatus = SEATOOL_TO_ONEMAC_STATUS[seaToolStatus];
+        putParams.Item.proposedEffectiveDate = DateTime.fromMillis(
+          anEvent.STATE_PLAN.PROPOSED_DATE
+        ).toFormat("yyyy-LL-dd");
         return;
       }
 
--- a/services/one-stream/package/buildAnyPackage.js
+++ b/services/one-stream/package/buildAnyPackage.js
@@ -142,10 +142,17 @@ export const buildAnyPackage = async (packageId, config) => {
           anEvent.STATE_PLAN.SPW_STATUS_ID,
           seaToolStatus
         );
-        putParams.Item.currentStatus = SEATOOL_TO_ONEMAC_STATUS[seaToolStatus];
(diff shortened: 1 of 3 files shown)


services/one-stream/package/buildAnyPackage.js
```diff
@@ -140,7 +142,17 @@ export const buildAnyPackage = async (packageId, config) => {
...
+          (putParams.Item.currentStatus =
+            SEATOOL_TO_ONEMAC_STATUS[seaToolStatus]);
+        if (
+          anEvent.STATE_PLAN.PROPOSED_DATE &&
+          typeof anEvent.STATE_PLAN.PROPOSED_DATE === "number"
+        )
+          putParams.Item.proposedEffectiveDate = DateTime.fromMillis(
+            anEvent.STATE_PLAN.PROPOSED_DATE
+          ).toFormat("yyyy-LL-dd");
+        else putParams.Item.proposedEffectiveDate = "none";
```
bflynn-collabralink, line 155: I know this was talked about but I wasnt clear on all the details. So if a record comes in with an empty, nonnumeric, or nonexistent PROPOSED_DATE field then we overwrite existing value and update to "none"?
I didnt catch the business reason but I guess there is a need to allow them to intentionally clear out the field?

bflynn-collabralink (approved):
Left a comment but otherwise makes sense to me!

Approved.

services/one-stream/package/buildAnyPackage.js
kristin-at-theta: so, I guess it does give them the ability to clear out the field... but the reason is that we can't be in the business of correcting SEA Tool - if they lose it or change it or clear it... that's on them :)

PR #1109, opened 2023-01-10
update formatPackageDetails

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22457

### Details

data.withdrawnByName and data.withdrawnByEmail were never being set/used. Therefore just remove this logic and only print a submittername if it is available. Verified new submisssions still include this data point but withdraw does not.

### Changes

- removed logic for including withdrawn by name and email from formatpackagedetails

### Test Plan

1. Login as statesubmitter
2. submit new package
3. verify email still contains proper information (including submitter name)
4. update status to In Review or locate another package that is available for withdraw
5. withdraw package and verify email does not include "undefined" name and email

### DIFF
--- a/services/app-api/email/formatPackageDetails.js
+++ b/services/app-api/email/formatPackageDetails.js
@@ -11,9 +11,9 @@ export const formatPackageDetails = (data, config) => {
       <p>
         <br><b>State or territory</b>: ${data.componentId.substring(0, 2)}
         ${
-          data.withdrawnByName
-            ? `<br><b>Withdrawn By</b>: ${data.withdrawnByName}<br><b>Email Address</b>: ${data.withdrawnByEmail}`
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
Interestingly enough.... I may have discovered what happened while I was testing.... because in my email account I clearly had withdrawal requests with the full package details!

I believe this is a consequence of the switch to using the stream to build the packages... before, we used a RETURN_ALL (or something like that) to return the package details after the update to the dynamodb record, so we DID have all the package details in the data.  By changing to the events... we lost that easy-lookup.

That relieves me a bit.... I couldn't figure out how we missed this when working on the withdraw request stuff and now I see :) 

Anyway, tested the emails, looks like everything works! The withdraw email is now quite sparse, but certainly they can go to the application to get those details and if the users really want it, guess we are adding that database lookup...

Approved!

PR #1114, opened 2023-01-13
FIXES: package builder too dependent on Action Types

kristin-at-theta:
### Story: PROD MIGRATION

### Details
updates to handle the Waiver RAIs broke the handling of the other packages.  Need to account for the field being present, but with a null value.

### Changes
- add a check for null in ACTIONTYPES
- reorganized the build switcher so that it does not trigger CodeQL findings
- sample json for createOneMacPackage updated to use the correct type format
- ignore events from the old SEATool topics State_Plan and RAI
- ONEMAC_STATUS.UNSUBMITTED did not exist... remove it from the mapping so the OneMAC status will take priority

### Test Plan
1. branch should deploy
2. All Component types (except for Temporary Extensions) should receive Status updates when their status is updated in SEA Tool
3. Waiver RAIs should still correctly map to the "closest" parent (based on ID parsing, so not always correct)

### DIFF
--- a/services/one-stream/handleOneStream.js
+++ b/services/one-stream/handleOneStream.js
@@ -52,8 +52,9 @@ export const main = async (eventBatch) => {
           case "SEATool": {
             const [, topic] = newEventData.GSI1pk.S.split("#");
             let actionType;
-            if (newEventData.ACTIONTYPES)
-              actionType = newEventData.ACTIONTYPES?.L.map((oneType) =>
+            console.log("%s ACTIONTYPES: ", inPK, newEventData.ACTIONTYPES);
+            if (!newEventData.ACTIONTYPES.NULL)
+              actionType = newEventData.ACTIONTYPES.L.map((oneType) =>
                 newEventData.STATE_PLAN.M.ACTION_TYPE.N ===
                 oneType.M.ACTION_ID.N
                   ? oneType.M.ACTION_NAME.S
--- a/services/one-stream/handleOneStream.js
+++ b/services/one-stream/handleOneStream.js
@@ -8,18 +8,6 @@ import { buildWaiverAmendment } from "./package/buildWaiverAmendment";
 import { buildWaiverAppendixK } from "./package/buildWaiverAppendixK";
 import { buildWaiverExtension } from "./package/buildWaiverExtension";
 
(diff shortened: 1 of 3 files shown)


services/one-stream/handleOneStream.js
```diff
@@ -139,14 +108,39 @@ export const main = async (eventBatch) => {
...
         }
-        if (typeof BUILD_PACKAGE_CALL[packageToBuild.type] === "function")
-          await BUILD_PACKAGE_CALL[packageToBuild.type](packageToBuild.id);
-        else
-          console.log(
-            `%s with type <%s> has no build package function??`,
-            inPK,
-            packageToBuild.type
-          );
+        switch (packageToBuild.type) {
```
bflynn-collabralink, line 111: This makes me sad that the tool forced us to break this out due to fake security concerns. :(

kristin-at-theta: Me too - I had flagged it as a "false positive" but the idea that it was in the reporting just kept nagging at me :(

services/one-stream/package/buildAnyPackage.js
```diff
@@ -15,7 +15,6 @@ const SEATOOL_TO_ONEMAC_STATUS = {
   [Workflow.SEATOOL_STATUS.TERMINATED]: Workflow.ONEMAC_STATUS.TERMINATED,
   [Workflow.SEATOOL_STATUS.PENDING_CONCURRENCE]:
     Workflow.ONEMAC_STATUS.PENDING_CONCURRENCE,
-  [Workflow.SEATOOL_STATUS.UNSUBMITTED]: Workflow.ONEMAC_STATUS.UNSUBMITTED,
```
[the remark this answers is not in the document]
kristin-at-theta: Yeah.... this is intentional... apparently SEA Tool has an Unsubmitted status, but it is no longer used and the spreadsheet of actions specifically called it out as unsupported... I had left in the conversion because I thought we might need it, but then, because Unsubmitted was not in the "approved status list" for the filters, the package was getting removed from the dashboard if it entered Unsubmitted... I figured the best solution was to remove it as an "Accepted status" so that the package status would not update and the package would still show on the dashboard with an older status.

PR #1116, opened 2023-01-18
Oy2 22443 - update excel export for package view

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22443

### Details

Fix package view export to excel

### Changes

- Use new column header names
- Use package view onemac type lookup instead of change request type lookup

### Test Plan

1. Login as system admin
2. Navigate to package view
3. export spa list
4. verify spa list export has new header value "Initial Submission Date" and Type of "Medicaid SPA"
5. export waiver list
6. verify waiver list export has new header value "Initial Submission Date" and Type of "Temporary Extension"


kristin-at-theta (approved):
Huh.... guess this script *is* split by dashboard :). So won't be all that bad if it accidentally gets released before the URL swap.

Code is clean, tests perfectly!  Approved :)

PR #1118, opened 2023-01-20
merge-sprint-56-57-58-59-to-production

Dark-Knight-000:

### Details

The purpose of this pull request is to merge sprints 56-59 ( including the data migration) from the validation environment to the production environment.

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2


services/admin/serverless.yml
```diff
@@ -36,6 +36,7 @@ provider:
             - dynamodb:UpdateItem
             - dynamodb:DeleteItem
           Resource:
+            - arn:aws:dynamodb:*:*:table/product-change-request-2
```
kristin-at-theta, line 39: This resource does not exist in the production account... which I was worried about until I noticed it does not exist in the master account (VAL) either, so probably not important.

kristin-at-theta (approved):
Yes, this looks correct !

Approved!

PR #1119, opened 2023-01-23
Oy2 22595 - always use seatool proposed effective date to backfill event data

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22595

### Details

Due to order of data processing when building packages, older ONEMAC events were being used to populate proposed effective date rather than the newer SEATOOL event.

### Changes

- Updated logic in buildAnyPackage

### Implementation Notes

- First attempt was to backfill all data from previous package record; however since ONEMAC records are processed first due to sk order the ONEMAC records were being used to populate proposed effective date and the newest status chane ONEMAC event had a later event timestamp than the SEATOOL event. Therefore the SEATOOL event was not used. For now since editing of a onemac package is not possible, assume any data other than status from SEATOOL can always be applied (currently only proposed effective date) regardless if its the latest event.

### Test Plan

1. Submit a package.
2. Submit a seatool event (or insert event into DB) to set a proposed effective date and PENDING status on the submitted package.
3. Withdraw package.
4. Verify pakcage retains proposed effective date set in seatool event.

### DIFF
--- a/services/one-stream/package/buildAnyPackage.js
+++ b/services/one-stream/package/buildAnyPackage.js
@@ -129,6 +129,16 @@ export const buildAnyPackage = async (packageId, config) => {
           lmTimestamp
         );
 
+        //always use seatool data to overwrite -- this will have to change if edit in onemac is allowed
+        if (
+          anEvent.STATE_PLAN.PROPOSED_DATE &&
+          typeof anEvent.STATE_PLAN.PROPOSED_DATE === "number"
+        )
+          putParams.Item.proposedEffectiveDate = DateTime.fromMillis(
+            anEvent.STATE_PLAN.PROPOSED_DATE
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
Thank you for diving into the event processing loop and seeing more than I did!  I agree with the choices you've made to perpetuate the SEA Tool proposed effective date.

Tested that the OneMAC proposed effective date is shown until there is a SEA Tool record, then the SEA Tool proposed effective date is shown, even if the SEA Tool event is older than the latest OneMAC event (so we see the OneMAC status and the SEA Tool proposed effective date).  Deleting the SEA Tool event reverts the proposed effective date to the OneMAC value (after a package rebuild, deletion does not trigger rebuild... )

Looks great, thanks !  Approved :)

PR #1121, opened 2023-01-23
OY2 22417: OneMAC should not distinguish between upper and lower case SEA Tool ID's.

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22417

### Details
SEA Tool has a level of case insensitivity that is unsupported by DynamoDB.  OneMAC upper-cases all IDs (per business requirements), but needs to also see the ones that have lower cased characters in SEA Tool.

### Changes
- Upper case the ID_NUMBER value when put into the pk and GSI1sk attributes.
- Removed old commented out mappings so we don't accidentally think they should be used.
- Upper case the ids sent in the createOneMacPackage lambda

### Test Plan
1. Login to AWS console
2. Navigate to logs for Lambda: seatool-sink-oy2-22417-seaToolEvent
3. Select a SEA Tool event to copy (I used a MD one in Pending-RAI status)
4. copy the event from the event logged
5. edit the event and use lower case letters in the ID_NUMBER field where ever it is (15 times in the one I used)
6. use the Test feature to submit the event to the lambda
7. Verify that the event was saved with the upper case pk and GSI1sk
8. Use Lambda: admin-oy2-22417-createOneMacPackage to create a OneMAC package for the ID (use some lower case) (make sure to put the Submission Date as a date BEFORE the SEA Tool event)
9. Login to the OneMAC application as a submitting user
10. Navigate to package view and look for the package
11. Verify details are correct
12. Verify actions taken on the package work as expected.

### DIFF
--- a/services/admin/handlers/createOneMacPackage.js
+++ b/services/admin/handlers/createOneMacPackage.js
@@ -31,17 +31,18 @@ export const createOneMacPackage = async (item) => {
   const submissionTimestamp = convertDateToSeaToolTimestamp(
     item.submissionDate
   );
+  const theID = item.componentId.toUpperCase();
   if (!item.territory)
     item.territory = item.componentId.toString().substring(0, 2);
   const putParams = {
     TableName: process.env.oneMacTableName,
     Item: {
-      pk: item.componentId,
+      pk: theID,
       sk: `OneMAC#${submissionTimestamp}`,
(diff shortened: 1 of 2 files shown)


bflynn-collabralink (approved):
Works well and if needed we should have the original id from the seatool event.

Approved!

PR #1127, opened 2023-01-26
remove override of actionLabel for amendments

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22594

### Details

Remove override of title for amendments. Keep default "Package Actions"

### Changes

- Removed "actionLabel" property from waiverAmendment config
- Removed redundant "actionLabel" property from waiverAppendixKDetail as it was same as in defaultDetail

### Test Plan

1. Login as any user with package view access
2. Navigate to any waiver package details page
3. Verify Label change


kristin-at-theta (approved):
Yes, the Package Actions label now reads "Package Actions" for Amendments!

I'm guessing it was the migrate-on-push feature that removed the Amendments you were looking at :P  Had to make my own?!

Approved :)

PR #1135, opened 2023-01-27
oy2-22685 - created new method to update packageID

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22685

### Details

Create new method to update a onemac package with a new package id

### Changes

- created updatePackageId

### Implementation Notes

- This new admin function will update all OneMAC events for a given package id
- Since we are updating the pk, we have to delete the original onemac event record and the package record then recreate the onemac event records (the new package record will be auto built by the event stream)
- only the original create event will get the additional info added (skip rai)
- a new field named auditArray to track updates has been added and is a simple string array with the prependAdditionalInfo data passed into this update method added as an element

### Test Plan

1. Login as a statesubmitter
2. Submit a new package
3. Run the lambda using the package as the "fromPackageId" property
4. Verify old package is gone and package with new id retained all properties with the updated additional information

### DIFF
--- a/services/admin/handlers/updatePackageId.js
+++ b/services/admin/handlers/updatePackageId.js
@@ -7,6 +7,7 @@ const dynamoDb = new AWS.DynamoDB.DocumentClient(
       }
     : {}
 );
+const oneMacTableName = process.env.oneMacTableName;
 
 function validateEvent(event) {
   //validate required input params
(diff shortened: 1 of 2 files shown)


kristin-at-theta (approved):
Code is clean and easy to read, per usual !

I feel like there is functionality missing... like what if you also want to convert the RAI Responses or what if the package is withdrawn or if there are other things changed in OneMAC... BUT... I feel like I need to squelch those thoughts because this script is being built for a specific need which does not include those things... so maybe it is better to wait until one of those situations is part of the ask before updating the scripts to handle them... 

So... Approved!

services/admin/handlers/updatePackageId.js
```diff
@@ -0,0 +1,146 @@
...
+  const queryResponse = await dynamoDb.query(params).promise();
+  console.log("OneMAC events found for: ", fromPackageId);
+
+  return queryResponse.Items;
+}
+
+function formatPutRequests(toPackageId, addlInfo, records) {
+  const putRequests = records.map((item) => {
+    const auditArray = item.auditArray
+      ? [...item.auditArray, addlInfo]
```
kristin-at-theta, line 52: SEA Tool's pattern is to prepend the most recent information to the front of whatever.... probably should have this the same direction...

bflynn-collabralink: updated this order in latest commit

bflynn-collabralink:
> Code is clean and easy to read, per usual !
> 
> I feel like there is functionality missing... like what if you also want to convert the RAI Responses or what if the package is withdrawn or if there are other things changed in OneMAC... BUT... I feel like I need to squelch those thoughts because this script is being built for a specific need which does not include those things... so maybe it is better to wait until one of those situations is part of the ask before updating the scripts to handle them...
> 
> So... Approved!

I think as is it "should" handle already withdrawn packages since its updating all OneMAC# event records, and it does work on a package that has rai responses, I did test that.
But I agree, we can dive in further when the use case is needed.

PR #1138, opened 2023-02-02
OY2 22729: Inactivated RAI Responses should be fully ignored

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22729
Endpoint: https://d21sxccb1vnkpk.cloudfront.net
(not sure why the github comment didn't work on this PR)

### Details
When Inactivating some RAI Responses, we discovered that the details of the Inactivated RAI Response were getting pulled into the package.  This is because the return statement was blocked by the status check.

### Changes
- moved the status check to Before the timestamp check (the newness of an inactivated event is irrelevant)
- included the softDeleteComponent function because building this function is how we found the bug and we could not use the function for the PROD data changes, so makes sense to include it here.
- softDeleteComponent function inactivates a component and updates the package accordingly
- updated the package builder to delete a package item if it should no longer be a package on the dashboard

### Test Plan
1. Login as a submitting user
2. Navigate to Package View
3. Find or create some components to inactivate, both package and RAI Response
4. login to AWS console
5. Use the softDeleteComponent lambda to inactivate the components
6. Verify the package details now ignore the inactivated components, either by not being on the dashboard, or by not being in the details.  For RAI Responses, the package details should not have any details left over from the inactivated RAI Response.
7. Verify that re-activating the components rebuilds the package properly


services/one-stream/package/buildAnyPackage.js
```diff
@@ -191,6 +192,22 @@ export const buildAnyPackage = async (packageId, config) => {
       putParams.Item.GSI1sk = packageId;
     } else {
       console.log("%s is not a OneMAC package: ", packageId, putParams.Item);
+      // if there was a package, and should not be now, delete the package item
```
bflynn-collabralink, line 195: Nice! Im not sure what the use case for this is outside of soft delete but it makes sense!

services/admin/handlers/softDeleteComponent.js
```diff
@@ -0,0 +1,85 @@
...
+
+  console.log("%s getParams", event.componentId, getParams);
+  const result = await dynamoDb.get(getParams).promise();
+
+  if (!result || !result.Item) {
+    throw new Error(pk + " not found with sk: " + sk);
+  }
+  const auditArray = result.Item.auditArray
+    ? [event.prependAdditionalInfo, ...result.Item?.auditArray]
+    : [event.prependAdditionalInfo];
```
bflynn-collabralink, line 68: Im happy to see this continued use of auditArray!

bflynn-collabralink (approved):
I like the soft delete changes in here too! Also I like how you incorporated the auditArray we created and kept that going!

soft delete looks good and rai updates look good as well!

Approved!

services/one-stream/package/buildAnyPackage.js
kristin-at-theta: I *think* it would also remove the package if the id got changed - I know you delete the package in the lambda you made, but in case someone just goes into the data and edits it, instead of using the lambda...

services/admin/handlers/softDeleteComponent.js
kristin-at-theta: :)

PR #1139, opened 2023-02-02
Passing95

Valencia2019:
Story: https://qmacbis.atlassian.net/browse/OY2-21942

### Details

Goal is to get automated tests passing at min 95% rate so we don't have to manually validate even more of the system. Some tests were outdated so they were updated. I also noticed there were some missing scenarios and took the opportunity to add new tests to cover those.


kristin-at-theta (approved):
Approved!

PR #1143, opened 2023-02-06
Oy2 21182

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-21182

### Details

Reroute /dashboard to the PackageList component.
Make current submission dashboard not navigable except via url.

### Changes

- Updated routes to make /dashboard route to packagelist components
- Created new route path so that /submissions would route to submission

### Test Plan

1. Login as any user
2. Verify landing page is the package list but url is /dashboard
3. Navigate to details page for any package and use the UI back button (not browser) and verify you go back to package list
4. submit a new package and verify you are returned to package list
5. login as a user without access to user management page such as a state submitter and manually type in the url /usermanagment. .....verify you are redirected back to package list


kristin-at-theta (approved):
While I do think the comment is perhaps a "more correct" way to go about doing this... experience has shown that these Users aren't that interested in a generic dashboard pattern.  What is done here works correctly and is consistent with the AC.

Nice work finding all the possible dashboard navigation places.

Approved!

Dark-Knight-000 (approved):
Cool. Note: Please merge to Dev after this evening's deployment

PR #1151, opened 2023-02-17
Update to API setup instructions

kevinhaube:
### Details

This PR updates the README in `services/app-api` with clarified and simplified setup instructions.

### Changes

- Remove redundant instructions calling for installs of packages already installed via `npm install` in the root and api directories
- Note clarifying the need for AWS credentials to proceed with the API setup process


kristin-at-theta (approved):
So... at first, I was all "oh, I should try the process to verify" and THEN I realized... "wait, that the NEXT NEW GUY/GAL's job :) " 

Reads well, the changes look like they add needed detail and remove outdated things!!

Approved :)

PR #1152, opened 2023-02-17
Update to UI setup instructions

kevinhaube:
### Details

This PR updates the setup instructions for the front-end directory, adding some clarity and removing outdated instructions.

### Changes

- No longer need to run `configureLocal.sh` since `env-config.js` is checked into Git and included in our repository
- Removed code block with test login user info in favor of a protected Confluence link
- Updated some descriptions and added callouts based on recent onboarding experience

### DIFF
--- a/services/ui-src/README.md
+++ b/services/ui-src/README.md
@@ -20,8 +20,6 @@ Running the start command will open [http://localhost:3000](http://localhost:300
 
 We have a suite of [OneMAC test users](https://qmacbis.atlassian.net/wiki/spaces/MACPRO/pages/2446852097/oneMAC+Test+Users) for the DEV environment that you can use for all feature branches and the `develop` branch, including when running the application locally. If you are unable to access the Confluence page, see a teammate for help!
 
-The password for all test users is `Passw0rd!`
-
 > **NOTE:** These users are not available for use in the `master` or `production` branch environments.
 
 ### Dynamic developer emails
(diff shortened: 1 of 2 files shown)


services/ui-src/README.md
```diff
@@ -1,56 +1,42 @@
...
-cmsroleapproverrevoked@cms.hhs.local
-cmsroleapproverunregistered@cms.hhs.local
-```
+### Static test users
 
-For every build, a dynamic list of developer emails is also added to the test user list, so developers can receive the generated emails sent from the application. Developers need to setup their correct email in the Git configuration and have already committed changes to the repo. To see the list of emails added run the following command in the Git repository.
+We have a suite of [OneMAC test users](https://qmacbis.atlassian.net/wiki/spaces/MACPRO/pages/2446852097/oneMAC+Test+Users) for the DEV environment that you can use for all feature branches and the `develop` branch, including when running the application locally. If you are unable to access the Confluence page, see a teammate for help!
 
-`git log --pretty=format:'%ae' | grep -v github.com | sort -u`
+> **NOTE:** These users are not available for use in the `master` or `production` branch environments.
```
kristin-at-theta, line 23: Actually, test users *are* available in VAL (master) now...

kristin-at-theta (approved):
Did not test the process as that is "next new devs" first job :) 

A small correction comment, but its not weighty enough for me to stop the...

Approved!

PS - what this means is that if you DO choose to update the text, it is automatically approved and I don't have to re-review.  It is also ok to not make the change and leave it for "next new dev"

services/ui-src/README.md
kevinhaube: Ooo, good to know! I'll fix it up 👍🏼

PR #1153, opened 2023-02-22
Oy2 22943

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22943

### Details

Create new cloudwatch alarm when no seatool event has been received after a given period of time.

### Changes

- Create new SNS topic
- Create new cloudwatch metric filter
- Create new cloudwatch alarm

### Implementation Notes

- The alarm will notify the SNS topic but there wont be any subscribers, we will want to subscribe the slack channel email to the VAL and Prod topics (and anywhere else desired)

### DIFF
--- a/services/seatool-sink/serverless.yml
+++ b/services/seatool-sink/serverless.yml
@@ -393,6 +393,7 @@ resources:
         MetricName: ConnectorEventReceivedCount
         AlarmActions:
           - Ref ConnectorMonitorTopic
+        ComparisonOperator: LessThanThreshold
         Statistic: Sum
         Period: 86400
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
I can't really figure out a way to test this other than to change what the alarm triggers on and see if stuff goes through... which I haven't the faintest idea how to do, so would have to research that... so... I did go into the console and I can see the Alarm, the SNS topic, the connections, and it looks like things are set up fine... so

Approved!

kristin-at-theta:
Ah - I do clearly see your tests in the onemac-core-team :)

PR #1155, opened 2023-02-24
OY2 22967: Update resetData lambda and one-seed

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22967

### Details

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2


services/admin/handlers/resetData.js
```diff
@@ -90,6 +143,37 @@ export const main = async (event) => {
...
+          ":inPk": id,
+        },
+        ProjectionExpression: "pk,sk",
+      };
+      try {
+        const results = await dynamoDb.query(qParams).promise();
+        console.log("Found these results in One Table: ", results);
+        for (const item of results.Items) {
+          const [, eventTimestamp] = item.sk.split("#");
+          if (eventTimestamp > 1676384054014)
```
bflynn-collabralink, line 162: I feel like i remember you saying something about using Feb14th as a cutoff date but I dont remember the reason! Assume it has to do with only wanting to delete test data created by automation that has that later timestamp but to keep some base records?

bflynn-collabralink (approved):
Looks good to me!

My only thought would be if we could cut out all the rebuilding of the application and call the lambdas direct via CLI since we have the deploy step finished by the time this executes. Same result just might reduce time and setup for this step. Food for thought for the future.

Approved!

PR #1156, opened 2023-02-27
Oy2 21488 - add readme pointing to confluence

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-21488

### Details

Added README.md to seatool-sink service
Updated REAME.md in onemac-stream service

### Test Plan
NONE. No code changes.


kristin-at-theta (approved):
Not a huge deal, but I was planning/expecting the seatool-sink to ONLY be SEA Tool Kafka topics and if/when we ingest from other topics, that would be a different sink.  Until we have a second sink, it's kinda irrelevant, but did want to mention it :) 

Approved!

PR #1159, opened 2023-03-02
Depend: uploads and admin breaking updates

kristin-at-theta:
Story: Dependabot

### Details
Updating packages and fixing breaks so we can get the dependabot calmed down.

### Changes
- Updated package.json in admin service to new serverless-offline version
- Updated package.json and serverless.yml in uploads service - removed local plugins because we don't use them?

### Test Plan
1. Branch should deploy
2. Application should run locally with no new trouble/issues
3. Application should run in AWS environment with no new issues


Dark-Knight-000 (approved):
Looks reasonable.

PR #1164, opened 2023-03-14
Save table filter and column visibility states

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23135

### Details

This PR updates the app so that it saves filter state and hidden column state when a user changes filter and column show/hide values on both the SPA and WAIVER tables independently. On reload of the page, the saved states are set as the default, so filtration and configuration can be saved in a browser instance for a user.

### Changes

- localStorage (read/write): `onemac-SPA-tableFiltersSavedState`, `onemac-SPA-columnHiddenSavedState`, `onemac-WAIVER-tableFiltersSavedState`, `onemac-WAIVER-columnHiddenSavedState`
- update `react-table.useTable` default table settings to include `hiddenColumns` and `filters` populated by the saved state or empty arrays if no state is saved
- removes `TEMP_onReset` as it was no longer used after implementing this feature

### Implementation Notes

Filters and hidden columns are all handled via `react-table` state inside the application, so none of our storage methods actually touch our filtering; every instance of the app may apply its own unique filters and column configs. All we are doing is extracting the most recently updated states and saving a copy to reload from when the browser comes back to the URL or refreshes the page for any reason.

Hidden Columns are stored as an array of strings, first in the tab's Session, then the same value overwrites the Local store. This is to prevent edits directly being made to a single source of truth from multiple tabs.

Filters are an array of objects, and because they're quite a bit easier to handle thanks to `react-table` managing that array for us, all filter changes do is overwrite the last object in the Local store, no Session needed.

#### More on the sessionStorage to localStorage thing

To ensure multiple tabs would not create a mutated source of truth in `localStorage`, we write any edited data, such as the array containing hidden column values, to `sessionStorage` in the particular tab, then use that edited value to overwrite whatever is in `localStorage`. The resulting UX is as such:

- Change filters in Tab A, switch to Tab B
- Refresh Tab B and see filters from Tab A applied
- Apply filters in Tab B, switch to Tab C
- Refresh Tab C and see filters from Tab B
- Change filters in Tab A, switch to Tab C
- Refresh Tab C and see filters from Tab A

In short, the most recent tab's saved filters and column show/hide states are the global source of truth for a reload in any tab.

### Test Plan

1. Use the UI to set up filters and column configs for the SPA table on your Dashboard
2. Change tabs to Waivers, and back to SPAs -- you should have persisted UI state including filtration and column setup
3. Refresh to ensure they load properly from storage
4. Change tabs to Waivers and repeat Step 1
5. Change tabs to SPAs, and back to Waivers -- you should have persisted UI state again
6. Repeat Step 3
7. Logout and back in
8. Navigate to dashboard and observe a reset of your filters and column setup


bflynn-collabralink (changes requested):
Nice work!

I do like that the updates persist across windows and even if you close and reopen your browser but remained loggged in....... as mentioned its ambiguous if that is the correct behavior or if the changes should be confined to the browser window the change was made in.

Another interesting thing I found is that if you make changes then they persist across both SPA and Waiver tables, again not clear in the AC if that should be the case or not but I would suspect you would at least want to have different configurations on different views within the application.  In the past we did have different columns in the two different data sets and even now there are different values that could be filtered between the two data sets. Though I do understand the complexity given the base table is the same actual table being populated with different data sets.

Finally, a definite bug is that  if you make selections on SPA tab then navigate directly to Waiver tab the selections are lost. Also navigating back to SPA tab they are still lost. So something with navigating tabs (staying on same url path) isnt saving the selection.

bflynn-collabralink (approved):
Both issues I noted have been cleared up.

Approved!

PR #1176, opened 2023-03-20
Oy2 22177 - excel export

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22177

### Details

Export only columns currently visible on screen.

### Changes

- Added useEffect in portal table to watch change in row data and send to new parameter function for outside consumption
- Updated package list 
-- store as state var  an array of current rows which is set by a parameter function into PortalTable
-- call new export function
- created new export function which is specific to portal table (react-table) and exports based off react-table `row` data structures

### Implementation Notes

- When creating the new export function I had to do something I didnt entirely like and this create another set of transform functions for data that is not held in the row the way we would like it to be. I tried using the method defined on the columns object of the table but this results in a JSX object being returned for rendering in a page and while this would have worked for our lone use case here (date formatting) it wouldnt necessarily work as a blanket operation. Therefore I created an map of id=>function to pass into export for it to transform any cell having that header id. Also the transformed value is wrapped in quotes to prevent commas in the data acting like seperate values in the resulting CSV file

### Test Plan

1. Login as state system admin
2. Navigate to package view (default view)
3. Export to Excel and verify all data is included
4. Hide some columns and filter some data to limit the displayed data set
5. Export to Excel and verify only data displayed is in the excel. Anything column hidden or row filtered out of view will not be in the report.

### DIFF
--- a/services/ui-src/src/components/PortalTable.tsx
+++ b/services/ui-src/src/components/PortalTable.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from "react";
+import React, { ReactNode, useEffect } from "react";
 import {
   HeaderGroup,
   TableInstance,
@@ -34,6 +34,7 @@ export type TableProps<V extends {}> = {
   searchBarTitle?: ReactNode;
   withSearchBar?: boolean;
   TEMP_onReset?: () => void;
+  onVisibleDataChange: (rows: Row<V>[]) => void;
 } & Pick<SearchFilterProps<V>, "pageContentRef"> &
   TableOptions<V>;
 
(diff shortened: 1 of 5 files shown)


kristin-at-theta (approved):
So... state system admin did not have an Export button, so I switched to the CMS system admin to test :) 

Code looks great, I like how the solution leverages the work the PortalTable has already done to create the data, even though you have to jump through some hoops to get it exposed at the right level for the button.

In testing, I tried downloading, show-hide columns, show-hide with filter, filter, and reset - and the data between the dashboard and the export matched each time.  I also verified that each tab maintains it's own state, so you get the data from the tab on which you clicked the button.

Nice work!  Approved!

PR #1195, opened 2023-03-31
Locks design system to 2.7.1 version

kevinhaube:
### Details

Fixes the design issues introduced after updating our design system package.

| Version | Active | Focused |
| ------- | ------- | -------- |
| Broken | Screen Shot 2023-03-31 at 8 40 32 AM | Screen Shot 2023-03-31 at 8 40 38 AM |
| Fixed | Screen Shot 2023-03-31 at 8 40 44 AM | Screen Shot 2023-03-31 at 8 40 48 AM

### Test Plan

Before testing, ensure you run `npm install --legacy-peer-deps` to update your package-lock.json

### DIFF
--- a/services/ui-src/package-lock.json
+++ b/services/ui-src/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@babel/polyfill": "^7.12.1",
-        "@cmsgov/design-system": "^2.7.0",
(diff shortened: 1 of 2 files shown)


kristin-at-theta (approved):
Nice find with the package craziness!  Also like how yoe are looking ahead to matching the next CMS Design System :) 

I can see the bug on Prod and it is not in the endpoint!
Approved :)

PR #1198, opened 2023-04-03
OY2 22141: Withdraw Package Functionality for State

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22141

### Details

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2


kristin-at-theta:
Thanks for that!  I looked to see if it was easy to update the < button landing page... but it looks like "not really". I think we originally used the url to assign the tabs, so the history would contain the tab setting when < was hit.  Now, we leave /dashboard in the URL, so there is no longer a tab history once you go to a form.

Basically, I verified that *all* waiver forms have the same issue, so feel like this is a general bug and not part of this feature.  Guess I will write up a bug ticket :P

PR #1200, opened 2023-04-03
Update Details Cards UI

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-22082

### Details

This PR:
- Extracts the `<DetailCard>` component for re-use
- Updates the UI on the Details page to show two `<DetailCard>`s (previously one unified card)

### Changes

- New file `DetailCard.tsx` contains extracted card component. This component is meant to be fed children and apply the card-like styling (i.e. the gradient-topped box) to whatever it is given.
- Updates to CSS for layout and scaling of cards
- Updates to JSX for layout

| Version | VAL | OY2-22082 |
| ----- | ----- | ----- |
| Desktop | Screen Shot 2023-04-03 at 12 31 53 PM | Screen Shot 2023-04-03 at 12 21 09 PM |
| Tablet | Screen Shot 2023-04-03 at 12 31 44 PM | Screen Shot 2023-04-03 at 12 21 28 PM |
| Mobile | Screen Shot 2023-04-03 at 12 31 37 PM | Screen Shot 2023-04-03 at 12 21 36 PM |

### Implementation Notes

- When possible, I'll be breaking out components like this to cut back on how much JSX we hand-write

### Test Plan

1. Pull latest and run `npm install` (for good measure)
2. Arrive at the details page of a SPA and/or Waiver by logging in as a test user and navigating to the dashboard, then clicking the hyperlink ID
3. Observe the new UI and ensure it matches the designs given
4. Use browser dev tools (right-click > inspect, or F12) to scale the size to view scaling styles on all screen sizes. Ensure no styling breaks


kristin-at-theta (changes requested):
I personally feel that making components out of style/layout similarities is a layer we don't need, but everyone has their preferences and I'll try to review as if I would also do that.

The components on the details page are fine.  I like how the added styles follow the pattern of the others.

The cards are well-formed and move nicely, I like how the margins adjust for the different sizes and that you include the formats you tested as part of the PR.

But if we are making the gradient top card a component, I feel like we need to convert *ALL* the places where that card is used, so we don't have two different ways of constructing the cards.  Some places I see:
- User Profile Access Management section
- FAQ Contact information
- the triage choice cards when clicking New Submission
- Home page "How tos"

kevinhaube:
@kristin-at-theta 

I decided against the new component because, after looking into those other cases, each card is using its own class for styling. If we were to want to do it "right" (that is, in a way that reduces our stylesheet and JSX), it'd greatly expand the scope of this work. I do love that idea, though, and think we should at least write a ticket up. I'll reach out to Mike Salamon so HCD has this in their ear.

kristin-at-theta (changes requested):
Well, phoo, that's disappointing!  I guess I should not be surprised that they are all different, each of those pages was developed by a different person - hopefully they at least all use Amy's gradient mixin!

I very much like the decision to stay light for now with an eye towards building something more comprehensive somewhere down the line.

This go 'round I compared more strictly to the Figma:
Figma has "No actions are currently available for this submission." in a size 16 font with weight 400 and app has that text in size 20 with no weight specified (is default 400?)

Card height is 128 in figma and 189.953 (!?) in application, creating more interior whitespace in application cards (top margin in Figma is 24 compared to 40 pixels of combined padding and margin in application.

By the way, I personally would approve even with those difference from the Figmas because I don't know which design details are important to make sure to keep exact and which should grow/change based on window, etc.  

However...  I also looked at a details page as a CMS user:
Screenshot 2023-04-03 at 3 16 17 PM
 
*that* should probably be fixed - sorry I missed it on the first go-round!

kevinhaube:
@kristin-at-theta It looks like it's because CMS users fail the `userRoleObj.canAccessForms` check -- I set it to default to showing that messaging rather than an empty box. Let me know if that has any implications for other user types I may have missed! :)

bflynn-collabralink (approved):
Warning, preference based comment follows! :)
I would rather see the userRoleObj.canAccessForms as part of an && statement with the action list size such that we dont have to define the section twice and duplicate that part of the jsx. Then no actions and no access to actions fall into the same logic group. Though this only works if we intend to keep the same language as we have now for no actions vs no allowed actions.

Having said that, I think all prior concerns have been addressed and it follows the ACs as defined.

Approved!

PR #1201, opened 2023-04-04
Oy2 22763 - remove all legacy functionality for change requests

bflynn-collabralink:
Story: https://qmacbis.atlassian.net/browse/OY2-22763

### Details

Remove any ui/api functions relating to submission and processing of change requests.

### Changes

- too many to enumerate but mostly removed all forms / routes / common libs / apis / IaC

### Implementation Notes

- Since we are removing the IaC templates in the serverless.yml, those legacy tables will be deleted upon deployment. Therefore we are copying legacy tables to s3 bucket before deployment so as to retain a copy that could be queried or rebuilt if needed.

### Test Plan

1. Verify application works as it did before removal


bflynn-collabralink:
Coverage dropped 1.2% just because of the sheer volume of code removed that apparently had higher test coverage average than the remaining code. I think we can "allow" this for now and build that back up as we move along since this PR isnt actually adding any new code.

services/ui-src/src/page/temporary-extension/TemporaryExtensionForm.test.js
```diff
@@ -7,12 +7,11 @@ import { stateSubmitterInitialAuthState } from "../../libs/testDataAppContext";
...
 import { ONEMAC_ROUTES } from "cmscommonlib";
 import TemporaryExtensionForm from "./TemporaryExtensionForm";
-import ChangeRequestDataApi from "../../utils/ChangeRequestDataApi";
+import PackageApi from "../../utils/PackageApi";
 import { AppContext } from "../../libs/contextLib";
 import { ONEMAC_TYPE } from "cmscommonlib/workflow";
-import PackageApi from "../../utils/PackageApi";
 
-jest.mock("../../utils/ChangeRequestDataApi");
+jest.mock("../../utils/PackageApi");
```
kristin-at-theta, line 14: Do we need to mock this twice?

kristin-at-theta (approved):
So this kind of "negative PR" is really difficult to review!

I can verify that the deletions here are correct... can't really verify if something was missed, though!

I checked that using the hidden urls for submission dashboard and the old forms give "page not found" responses.  

I looked in API Gateway through the AWS console, and do not see any changeRequest api calls

The only table for that stage is the one table

I trust the running of the application to the tests... I see one test failing, but that test is currently failing on develop branch and looks to be something with the seed data?  anyway, not part of this PR.

Approved!

PR #1206, opened 2023-04-10
Fix `onemac-form` flex direction and alignment

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23406

### Details

Adjusts styles in the `onemac-form` class used in the OneMacForm component so that the two items in the flex box are in a single column at all screen sizes.

### Changes

- Flex direction was row, now is column
- Added center alignment to fix positioning once flex direction was changed

### Implementation Notes

- N/A

### Test Plan

1. Log in using a dev account
2. Navigate to the Dashboard
3. Click the "New Package" button
4. Ensure each form can scale to a 4K or higher resolution without breaking styles (use browser dev tools or CMD/CTRL and `-` to downsize the screen in your browser to emulate higher resolutions.

### DIFF
--- a/services/ui-src/src/index.scss
+++ b/services/ui-src/src/index.scss
@@ -590,6 +590,8 @@ input[type="file"] {
   @extend .app-bar;
   @extend .ds-u-padding-top--5;
   @extend .ds-u-justify-content--center;
+  flex-direction: column;
+  align-items: center;
 
   form {
     @include max-width-ultra-wide;


kristin-at-theta (approved):
Tests correctly - thank you very much for the instructions on how to show the bug on my gimpy, low-rez system :) 

I don't get why **@extend .ds-u-justify-content--center;** is not the same as **align-items: center;** but... I'm not upset by my ignorance here :) !

Approved!

kevinhaube:
@kristin-at-theta From my extrapolation, justify center can help achieve the centered result for a row in horizontal space, and flipping it to a column means its "centering" justification in vertical space. Likewise, using align-items in a column direction aligns them along a horizontal axis, but when you go back to rows, you can align all items to the top, bottom, sides, or center of your vertical flex box space. :) 

But agreed: unintuitive at times!

PR #1207, opened 2023-04-10
OY2 23115: Dashboard Updates: Default Settings (hide columns based on role)

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23115

### Details
In order to keep the dashboard table from being too wide, certain columns are going to be hidden as a default.  This is based on role (currently State v CMS).

### Changes
- created a default hidden array for state and cms
- mapped the default hidden arrays to the roles
- put the default as the "default" if there is nothing in the session/local variables.

### Test Plan
1. Login as State user
2. Verify that "State" column defaults to hidden
3. Verify that "State" column can be added/removed as expected
4. Verify that no other functionality has changed with saving/showing the columns
5. Login as CMS user
6. Verify that "Submitted by" column defaults to hidden
7. Verify that "Submitted by" column can be added/removed as expected
8. Verify that no other functionality has changed with saving/showing the columns


bflynn-collabralink (approved):
Looks good! I like how you mapped each role to a set of defaults! I think it also works doubly since if a new role were added but we missed adding to this array then the array lookup would return `undefined` and still have no hidden defaults but not break the code....I think. :)

Approved!

PR #1218, opened 2023-04-14
OY2 23777: Adjust Lambda timeout value to prevent loss of email notifications.

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23777

### Details
Form submission lambdas use the standard lambda timeout of 6 seconds for processing.  This is plenty of time for most operations, but there were 2 times when sufficient load on the dynamodb table resulted in a timeout.  Analyzing the logs suggests a timeout of 15 seconds should be sufficient, so might as well go for 30!  Care should be taken to observe if increasing the timeout has side effects.

### Changes
- added `timeout: 30` to any lambda with a database-to-email dependency

### Test Plan
1. Successful Deployment
2. Verify all lambdas with database-to-email dependencies have the new timeout value.
3. Login as a submitting user
4. Navigate to any new submission
5. Submit a new Medicaid SPA, CHIP SPA, or waiver
6. Verify the lambda completes
7. Verify the email is sent (can be done through console)

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -98,6 +98,7 @@ functions:
   submitInitialWaiver:
     handler: form/submitInitialWaiver.main
     role: LambdaApiRole
+    timeout: 30
     events:
       - http:
           path: submitInitialWaiver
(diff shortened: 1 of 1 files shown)


bflynn-collabralink (approved):
Makes sense to me, I dont think there would be any negative consequences other than longer running lambda and therefore a very small cost. However as we expect this to only be occasional i still think 30sec is reasonable.

Approved!

PR #1221, opened 2023-04-18
Oy2 22431 - withdraw banner updates and post navigation

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-22431

### Details

Update banner text after succesful withdraw submission. After submission navigate user back to origin (package list or details) based on where they initiated the withdraw request.

### Changes

- Added a new RESPONSE_CODE for withdraw requested
- Added new message mapping for withdraw requested RESPONSE_CODE
- Updated withdraw success to use new RESPONSE_CODE
- Added new domain type for formSource
- Added formSource as property of action component
- Withdraw action now uses formSource to add to location.state
- OneMacForm looks at location.state and if a formSource exists it will override the configured landingPage

### Implementation Notes

- I only enabled the withdraw action to use formSource; but any action should be able to now use it with minimal changes
- Since I already built a mechanism to track where a user is coming from when initiating a form submission; I also applied this to fix the back navigation in the header so that the user will be redirected to the proper package list Tab based on the form configuration

### Test Plan

1. Login as statesubmitter
2. Locate a package that has a withdraw action
3. From the package list use the elipsis action button to select the Withdraw Package action
4. Submit the withdraw form
5. Verify user is returned to package list and banner message is reflective of ACs
6. Find another package that has the withdraw package action available
7. Navigate to the package details page and click the Withdraw Package action
8. Submit the withdraw form
9. Verify user is returned to the package details page and banner message is reflective of ACs


kristin-at-theta (approved):
Looks like it correctly returned user to the calling page and also has the new messaging!  I think I probably would've cheated and used the history, so nice to see you made it a real destination configuration :) 

Approved!

PR #1222, opened 2023-04-18
override chip spa to address with proper chip spa env var

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23697

### Details

Chip spa emails are not being sent to correct primary distribution list

### Changes

- override default cms email with chip specific email (from env var)

### Test Plan

1. Set reviewerChipEmail env var on submit lambda
2. Submit new chip spa
3. verify email


kristin-at-theta (approved):
So... I feel like what you have done works and is certainly a light code lift.  I guess I feel like why are we carrying the two variables around when we have separated the lambdas?  IMO, would be "cleaner" to keep using reviewerEmail for all the submits and change the value for the CHIP forms.... though I suppose I don't know how that translates into "infrastructure as code" model... but not sure this really does, either... more reason to have an admin!

Also, do the CHIP SPA RAI and CHIP SPA Withdraw forms also need to go to the CHIP email address?  I tested the RAI form and did not receive an email.  I see that the story does not include those forms in the AC... 

Approved

bflynn-collabralink:
> So... I feel like what you have done works and is certainly a light code lift. I guess I feel like why are we carrying the two variables around when we have separated the lambdas? IMO, would be "cleaner" to keep using reviewerEmail for all the submits and change the value for the CHIP forms.... though I suppose I don't know how that translates into "infrastructure as code" model... but not sure this really does, either... more reason to have an admin!
> 
> Also, do the CHIP SPA RAI and CHIP SPA Withdraw forms also need to go to the CHIP email address? I tested the RAI form and did not receive an email. I see that the story does not include those forms in the AC...
> 
> Approved

I agree. I should at minimum add the override on the other CHIP forms. I didnt think about splitting vars by lambda which we should totally do! I will consider doing that as well.

PR #1223, opened 2023-04-18
OY2 22615: Update CMS Package Details pages with Subject + Description fields from SEA Tool

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-22615

### Details
CMS Users (minus CMS Role Approver and add HelpDesk) would like to see two addtional fields from SEA Tool: Subject and Description.

### Changes
- create a canSeeSubjectAndDescription flag because who can see it is unique...
- getDetail checks to see if it should send that information, if not, it removes it from the result item. (probably want to go the other way at some point.... we have the list of information each role is allowed to see... so we don't inadvertantly send information to the front end that is not supposed to be there)
- modified the buildAnyPackage to include those two fields from the latest SEA Tool event when building packages
- Updated the details pages to include the subject and description, if included

### Implementation Notes
- Because this branch modifies how the build package script builds every package, we needed a way to cycle through all the packages.

### Test Plan
1. Log in as a user with Subject and Description privileges: CMS Read-Only, CMS System Admin, Helpdesk
2. Click to get details on a package that clearly has a SEA Tool record (status other than Submitted - Intake needed)
3. View the Subject and Description fields in the details
4. Logout
5. Login as a user without Subject and Description priviledges
6. Click to get details on the same package
7. Verify that the Subject and Description are NOT visible.


services/ui-src/src/libs/detailLib.ts
```diff
@@ -100,15 +108,13 @@ export const defaultDetailSectionItems = [
...
 ];
 
 export const defaultWaiverDetailSectionItems = [
   waiverAuthorityDefault,
-  typeDefault,
-  territoryDefault,
-  submissionDateDefault,
-  latestRaiResponseDateDefault,
-  proposedEffectiveDateDefault,
+  ...defaultDetailSectionItems,
```
bflynn-cms, line 117: nice!

bflynn-cms (approved):
Everything codewise looks good to me! Will be interested how that rebuild packages integrates with things.
Tests show the expected users can see the field and state users cannot and that TEs do not show the fields even for CMS users.
ACs seemed a bit fuzzy on placement so I think what you chose makes the most sense given the limited information in the story.

Also nice sneaking in package updates in there!

Approved!

PR #1227, opened 2023-04-20
OY2-23685: Pass empty array to DateFilter on reset

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23685

### Details

This PR fixes a bug where the DateFilter UI would not reset when the filter panel's Reset button was pressed/clicked.

### Changes

- Adds a special return for DateFilter resetting that passes `[]` instead of `undefined`

### Implementation Notes

- Thanks to the dev who left the comment! I updated the comment to give some more context! Good luck to the next debugger!

### Test Plan

1. Log in and navigate to the filter panel on the Dashboard page
2. Use the date filter to set a range for the `Initial Submission` filter
3. Use the reset button to clear this filter.
4. Repeat steps 2-3 a few times using multiple options like "Last 7 Days", etc. to ensure this works across the board! I saw it was hit-or-miss in VAL, but this should work unanimously.


kristin-at-theta (approved):
Ok, I am not the best at git... BUT... where did this get lost?  I can clearly see the code missing from the develop branch, but looking at the history shows that it was in the code for the last commit on this file?!  I am very confused...

Anyway, this fix is Approved!

PR #1229, opened 2023-04-25
OY2-23614 - let state users view latest rai response date on dashboard

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23614

### Details

Add Rai Response to state user dashboard. The column header name varies slightly from the column header for CMS users.

### Changes

### Implementation Notes

- Data for RAI response was already being returned for all user types. Just need to add it as a column element for state users.

### Test Plan

1. Login as state submitter
2. On Package list verify new column header "Formal RAI Response"
3. Locate a package with an RAI response and verify the value in the column matches the value on the details page


services/ui-src/src/containers/PackageList.js
```diff
@@ -233,7 +233,14 @@ const PackageList = () => {
               filter: CustomFilterTypes.DateRange,
               Filter: CustomFilterUi.DateRangeInPast,
             }
-          : null,
+          : {
```
kristin-at-theta, line 236: I probably would have moved the isCMSUser check to the Header, since all the rest is the same..... but this is less of a code change and wouldn't interfere with what was done for CMS users, so this way is not a bad choice, either.

kristin-at-theta (approved):
tests correctly, verified the column name difference between CMS and State users.

Can sort and filter on the date, too.

Helpdesk user see the same name as the State, which matches that Helpdesk sees the state values for status (but they do see the subject and description like a CMS user).  Nothing to change for that, I think the right choice was made, given there was no guidance in the story.

Approved!

PR #1233, opened 2023-04-27
Fix to CloudFront logging bucket, courtesy of macpro-mdct-mcr repo

kristin-at-theta:
This should work... though I had to run deploy twice because it said it failed on ui-auth... but when I looked in the AWS console, the stack was there and complete, so I re-ran deploy and everything's green!


bflynn-cms (approved):
Interesting. Im glad you found this!

PR #1238, opened 2023-05-01
Fix overflow for actions menu

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23749

### Details

Removes unnecessary (and breaking) overflow-x control in style class to fix the bug.

### Changes

- Change `overflow-x` control for `dashboard-and-alert-bar`:  was `auto`, now `scroll`. Just a more declarative way of styling is all. Functionality should remain the same
- Turn `#root` from a flex-box column to a flex-box column _with `space-between` justification_. This makes the headers and footers stick to the top and bottom of a viewport while the middle can be its own content container.
- Add `padding-bottom` to the table's container to allow for y-axis pop-up menu space without scrolling. Because of its need to be `absolute` in its positioning `relative` to its parent, we cannot rely on `overflow-y` controls.

#### Before
Screen Shot 2023-05-02 at 3 13 51 PM
Screen Shot 2023-05-02 at 3 14 04 PM

#### After

Screen Shot 2023-05-02 at 3 08 34 PM
Screen Shot 2023-05-02 at 3 09 05 PM

### Test Plan

1. Grab any ID on a Dashboard row that has package actions available in the 3-dot menu
2. Enter the exact ID in search to ensure it's the only item on the table
3. Click the 3-dot menu under Actions column
4. Ensure the fixed styles are represented ("After" above)


bflynn-cms:
I think this might be breaking the flow at 1024x768 resolution. Previously we had made things work at that resolution and up. If we no longer care about this size then we should note that.
The behaivor im seeing is on develop the whole table "fits" within the 1024 width though it does have an interesting vertical scroll bar for a few pixels:
image

However in this feature branch it totally breaks and causes the table to expand outside the width of the headers.
image

This used to function using breakpoints to keep the table within the 1024 until it couldnt fit then it would expand from the center but it appears recent changes have disrupted all that. Obviously old was was still flawed as well given this overflow of the popup being hidden but I think we need to find a way to make everything work at the same. Maybe the solution in this PR would be ok if the header/footer also expanded to fit the space?

kristin-at-theta:
I've kinda wondered if the best fix is to pad the table at the bottom, so there's room for the menu.  It's not elegant, but Brandon is correct... that whole page is fine-tuned and clipping was determined to be the least of the evils....

kevinhaube:
The original issue with just padding the table was the way the whole `#root` element lays itself out. We are not justifying the content meaning headers and footers will remain at the top and bottom rather than needing those `flex` grow and shrink properties to try and pin it to the bottom. Now our "middle box" of the 3-box `#root` will always be a container for all content and can scroll on the x-axis independent of anything else, as can its sub-boxes.

Ready for re-review @kristin-at-theta @bflynn-cms

bflynn-cms (approved):
Things look much better in regards to the scroll situation between header and table. Thanks for fixing that!

Approved.

PR #1240, opened 2023-05-05
OY2 23720: Update OneMAC packages from SEATool Delete notifications

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23720

### Details
When a package is deleted from SEA Tool, we get a null event for that package.  OneMAC needs to read that event and perform the delete action.
Deleting means that the package is no longer visible in OneMAC dashboards, and the ID is reusable.
Auditing requires that we keep the package details as well as the information about the deletion.

### Changes
- updated the seatool-sink to recognize and flag a delete event from SEA Tool
- updated handleOneStream to recognize the delete flag and call the delete function
- created deletePackage function which takes a package ID and pushes all events for that ID into the 

### Implementation Notes
- 

### Test Plan
1. Test step 1
2. Test step 2

### DIFF
--- a/services/one-stream/package/deletePackage.js
+++ b/services/one-stream/package/deletePackage.js
@@ -1,8 +1,6 @@
-const _ = require("lodash");
 import AWS from "aws-sdk";
-import { DateTime } from "luxon";
 
-import { dynamoConfig, Workflow } from "cmscommonlib";
+import { dynamoConfig } from "cmscommonlib";
 
 const dynamoDb = new AWS.DynamoDB.DocumentClient(dynamoConfig);
 
@@ -35,7 +33,7 @@ export const deletePackage = async (packageId) => {
           Item: {
             ...item,
             pk: "DELETED_ITEMS",
-            sk: `${pk}#${item.sk}`,
+            sk: `${item.pk}#${item.sk}`,
           },
         };
         await dynamoDb.put(putParams).promise();
--- a/services/one-stream/package/deletePackage.js
+++ b/services/one-stream/package/deletePackage.js
@@ -1,0 +1,62 @@
+const _ = require("lodash");
+import AWS from "aws-sdk";
+import { DateTime } from "luxon";
+
+import { dynamoConfig, Workflow } from "cmscommonlib";
+
+const dynamoDb = new AWS.DynamoDB.DocumentClient(dynamoConfig);
+
+const oneMacTableName = process.env.IS_OFFLINE
+  ? process.env.localTableName
+  : process.env.oneMacTableName;
+
+export const deletePackage = async (packageId) => {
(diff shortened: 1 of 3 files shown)


services/one-stream/package/deletePackage.js
```diff
@@ -0,0 +1,64 @@
...
+    if (result?.Items.length <= 0) {
+      console.log("%s did not have Items?", packageId);
+      return;
+    }
+    await Promise.all(
+      result.Items.map(async (item) => {
+        if (item.GSI1pk) delete item.GSI1pk;
+        if (item.GSI1sk) delete item.GSI1sk;
+        if (item.GSI2pk) delete item.GSI2pk;
+        if (item.GSI2sk) delete item.GSI2sk;
```
bflynn-cms, line 34: doesnt make any difference but just noted that we dont need to test for property existence before invoking delete, if the property doesnt exist the delete operation would just ignore and continue

services/seatool-sink/handlers/seaToolEvent.js
```diff
@@ -11,18 +11,19 @@ function myHandler(event) {
...
   const topicName = event.topic.replace(topicPrefix, "");
-  const value = JSON.parse(event.value);
+  const value = event?.value ? JSON.parse(event.value) : event;
+  if (!event?.value) value.seaToolDelete = "true";
   console.log(`Type: ${topicName} Topic: ${event.topic}`);
 
-  const pk = value.STATE_PLAN.ID_NUMBER.toUpperCase();
-  const changedDate = value.STATE_PLAN.CHANGED_DATE;
+  const pk = JSON.parse(event.key).toUpperCase();
+  const changedDate = value?.STATE_PLAN?.CHANGED_DATE || value.timestamp;
```
bflynn-cms, line 19: I do wonder what the downstream effects of using value.timestamp in lieu of STATE_PLAN.CHANGED_DATE would be as far as if we could get a real SEATOOL event (non delete) that is missing the STATE_PLAN.CHANGED_DATE and we should have bailed (thrown an error even) rather than using the event timestamp.

Thats probably not likely and would be the source data fault.....id just be worried about accidentally processing something without us realizing the data fault. But perhaps im reading this wrong too.

bflynn-cms (approved):
I dont have seatool access but the test provided made sense given the code changes.
Made a few non issue comments along the way.

Approved!

services/one-stream/package/deletePackage.js
kristin-at-theta: Huh, I could've sworn I got exceptions if the attribute didn't exist, but I do sometimes try to predict where an exception could happen, too.... I'll have to try it sometime :)

Dark-Knight-1313 (approved):
I have my access back WooHoo.

PR #1242, opened 2023-05-09
Oy2 23917 - add new forms and config for submitting 1915b waivers

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23917

### Details

Add different forms and config for each of the 1915b waiver types so that they are seperated by waiverAuthority. Each waiver type and waiver authority type combination has a different set of required and optional attachments.

### Changes

- Split each of the waiver forms (Iniitial, Renewal, Amendment) into 2 forms, one for (b)(4) and one for b other.
- Adapted submitXXXXX endpoints for the 3 waiver types to determine configuration for validation and submittal based on the submitted waiver authority

### Implementation Notes

- I didnt go as far as creating new "types" since that would require far more extensive changes as well as some UI impacts that were not part of the story. This is why I altered the api submission classes to determine the proper config rather than creating new submission endpoints as the submission endpoints work largely off of submission type

### Test Plan

1. Login as state submitter
2. Submit one each of the 6 new forms - for each one validate required attachments match the ACs
3. Progress a new waiver submission to RAI Issued status and verify you can still submit an RAI response
4. Progress a new waiver submission to In Review and verify you can still withdraw


kevinhaube (commented):
Approving from the front-end purview! 👍🏼 Still not confident singing off on back-end code standards so I'll rely on Kristin's review for that!

Also BONUS -- shouldn't be any merge conflicts when I pull this in :D

services/common/type/waiverAmendment.js
```diff
@@ -44,3 +30,22 @@ export const waiverAmendment = {
     "parentType",
   ],
 };
+
+export const waiverAmendmentB4 = {
+  ...waiverAmendment,
+  waiverAuthority: waiverAuthorityB4,
+  requiredAttachments: [
+    "1915(b)(4) FFS Selective Contracting (Streamlined) waiver application pre-print (Initial, Renewal, Amendment)",
```
kristin-at-theta, line 38: The AC does not include "(Initial, Renewal, Amendment)" for this attachment

kristin-at-theta (changes requested):
Things not in AC, but I noticed:
Authority is not checked for parent ID in the form: i.e., when in the Other 1915(b) waiver form, I am allowed to put a waiver number that is from the 1915(b)(4) authority and vice-versa.

some places don't exactly match the spreadsheet... though the spreadsheet is also all over the place with capitalization, etc.  It does look like some of the text replacements from 24006 were made, but not everywhere...

1915(b) Other has:
1915(b) Comprehensive (Capitated) Waiver Cost effectiveness spreadsheets
Spreadsheet has:
1915(b) Comprehensive Cost effectiveness spreadsheets 

Change requested: Please fix the attachment type that still has the (Initial, Renewal, Amendment) text in it.

kristin-at-theta (approved):

PR #1243, opened 2023-05-09
Add FAQ section and links from Attachments on form pages

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23750

### Details

This PR adds a new FAQ section about the various file types we allow for attachments. It also adds a link to that section along with some copy provided. Finally, it customizes the FAQ link that already exists to allow for routing to specific sections of the FAQ page (i.e. we link directly to that form's relevant attachment descriptions on the FAQ page). 

### Changes

- Upgrades to the components handling display of this UI (also enhances uniformity)
- New FAQ section added
- Updated ROUTES with form attachment related anchors

### Implementation Notes

N/A

### Test Plan

1. Using the New Submission button, navigate through the card options to a form
2. In the attachments section of that form, ensure the copy meets the requirements within the Jira ticket
3. Ensure the first FAQ link meets the requirement of navigating to the attachments FAQ section for the given form
4. Ensure the second FAQ link meets the requirement of navigating to the file types FAQ section
5. Repeat 1-4 with the form pages given in the Jira ticket's AC


services/ui-src/src/utils/fileTypes.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+
+interface FileTypeInfo {
+  extension: string;
+  description: string;
+}
+/** The source of truth for filetypes accepted as attachments in forms. */
+export const FILE_TYPES: FileTypeInfo[] = [
```
kristin-at-theta, line 8: This list does not match the allow list from serverless.yml in uploads directory.  I see this is what the Figma has, but there is absolutely NO WAY we are allowing .dmg files to be uploaded without explicit instructions to do so.

kevinhaube: HA! Yes, `.dmg` had me _so curious_! I was like "idk, maybe it's a gov sec thing" 😂 I'll reference the `serverless.yml` -- we have a few of those right; which directory has that specific one?

kevinhaube:
@kristin-at-theta Just a follow-up on the `_blank` and `_new` thing -- because I was curious -- I found this helpful! https://stackoverflow.com/questions/4964130/target-blank-vs-target-new

services/ui-src/src/libs/formLib.tsx
```diff
@@ -45,24 +47,57 @@ type ConfirmSubmitType = {
...
+export const DefaultFileTypesInfo = () => (
+  <p>
+    We accept the following file types:{" "}
+    <b>
+      .doc, .docx, .jpg, .odp, .ods, .odt, .png, .pdf, .ppt, .pptx, .rtf, .txt,
+      .xls, .xlsx, and a few others.
+    </b>{" "}
+    See the full list on the{" "}
+    <Link to={ROUTES.FAQ_ACCEPTED_FILE_TYPES} target="_blank">
+      FAQ Page
```
kristin-at-theta, line 59: I think I wasn't clear on the target="_blank" versus target="New" request - In my opinion, they should all be the same, either all new windows or the singleton window named "new" - right now, there are five links to the FAQ page on a form page, 2 go to "New" and 3 open a new window.

kristin-at-theta (changes requested):
just call me the nit-picker :)

bflynn-cms (changes requested):
Not sure if this was ready for another review or not but just wanted to note the following discrepancies still exist:

- "Maximum file size of 80 MB." should be "Maximum file size of 80 MB per attachment."
- "* indicates an attachment is required." should be "* indicates required attachment."

Other than that I verified rest of the text and that FAQ links all go to appropriate FAQ section.

services/ui-src/src/libs/formLib.tsx
kevinhaube: None of them open a new window on my end, everything opens a new tab only. If we want to file the change for other links as a tech debt 1-pointer we can, but I fail to see how this relates to the AC of _this_ ticket. [According to the HTML5 spec](https://html.spec.whatwg.org/multipage/document-sequences.html#browsing-context-names), we should be using `_blank`.

kristin-at-theta: I have very little opinion on what way they work, I personally would like all FAQ links on a page to work the same - which they do not do in my Chrome browser.  But this is not in the AC and there is certainly room for disagreement, so I will retract my request for this change.

services/ui-src/src/libs/formLib.tsx
```diff
@@ -45,24 +47,57 @@ type ConfirmSubmitType = {
...
+    </Link>
+    .
+  </p>
+);
+
+export const RequiredAttachmentSpan = () => (
+  <p className="req-message">
+    <span className="required-mark">*</span> At least one attachment is
+    required.
```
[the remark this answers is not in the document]
kristin-at-theta: So... now we are seeing * indicates required attachment on the withdraw package forms, of which, only the CHIP SPA has a required attachment.....
Screenshot 2023-05-23 at 11 19 42 AM

kristin-at-theta (changes requested):
Screenshot 2023-05-23 at 11 19 42 AM
Sorry to have missed this before... but the changes made to the Attachment text have bled into the Withdraw forms, which have their own attachment text specific to their requirements.

services/ui-src/src/libs/formLib.tsx
kevinhaube: I think I'm gonna demo from this PR's environment link, and pose the question so we know from the top what the expected behavior is (or at least convince them to go one way or the other if they don't have an opinion) because they should be uniform, and I think they'll have a preference even if we don't. Like I said, quick change and it'd be good to go, but I'm kinda in a "_always_ have a requirement first" state of mind due to the vague ACs lately.

kevinhaube:
@kristin-at-theta Haha, oops, I got lost in the forms and took a wrong turn apparently -- removed the generic one and re-added the bespoke text from before

kristin-at-theta:
Getting lost in the forms is an every day event for me :) 

Anyway, I see the Required text is not on the Withdraw forms (finally.... took FOREVER to deploy, ok, maybe just seemed that way)

I think that is everything!  approving changes :)

kristin-at-theta (approved):

bflynn-cms (approved):

PR #1244, opened 2023-05-11
Seed-Data: Add subject and description to seed data packages

kristin-at-theta:
Story: none, tech debt

### Details
Because changes we make to the data structures in our database need to go up the environments, we create lambdas to migrate the data.  However, we do not want to keep having to run those migrations for every new feature branch, so we have to update the seed data so that new branches have the new format for the data.  The update should happen *after* the migration has been launched up through production.

### Changes
- Removed some extraneous items
- added  
```
"subject": "test subject",
"description": "test description",
```
to every package item

### Test Plan
1. Branch should successfully deploy
2. There should be no automated test errors related to missing Subject and Descriptions


bflynn-cms (approved):
Thanks for explaining the need for this approach!

Approved!

bflynn-cms (approved):
Thanks for explaining the need for this approach!

Approved!

PR #1248, opened 2023-05-16
OY2 23424: TE Submissions Issues - Refactor

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23424

### Details
We missed validating the parent IDs using the SEA Tool data (with no corresponding OneMAC submission) - the original fix dug into the SEA Tool events to do the matching, but that puts SEA Tool business logic into app-api.  We decided it was better to contain all SEA Tool logic in the seatool-sink service.  That means that all packages have to have package items, though only the ones marked with the GSI1pk for the tabs will show in the tabs.

### Changes
- create a "Package" item for each ID known to the system
- limit the use of GSI1pk "OneMAC#spa" and "OneMAC#waiver" to packages that should be shown on OneMAC dashboard.
- validating parent now only needs to search "Package" items
- optimized rebuild packages so that we did not hit throttling limits

### Implementation Notes
Throttling on rebuild packages is really rather bad, as AWS tends to "drop" events that come in after a throttle... this is causing major data integrity issues, fortunately only in the building of packages needs.

### Test Plan
1. Use AWS Console to find a valid and an invalid parent package (right type and status) that is not a OneMAC package (does not have GSI1pk)
2. Application should deploy
3. Automated tests should run
4. Login as a submitter user
5. Navigate to a form where your parent ID can be used
6. verify that the form allows the valid parent
7. verify that the form does not allow the invalid parent


services/one-stream/package/buildAnyPackage.js
```diff
@@ -211,23 +211,6 @@ export const buildAnyPackage = async (packageId, config) => {
       putParams.Item.GSI1sk = packageId;
     } else {
       console.log("%s is not a OneMAC package: ", packageId, putParams.Item);
-      // if there was a package, and should not be now, delete the package item
-      if (currentPackage) {
-        const deleteParams = {
```
bflynn-cms, line 216: we might have relied on this behavior in admin functions for when we did a change id
this is just a note to self to double check on this and if any changes would be needed in that regard.

bflynn-cms: checked this and we do the delete of the package in the change id function and do not rely upon this here, which is good!

bflynn-cms (approved):
I admit that at first I thought it odd to create Package items for seatool events....but it does make the lookup much cleaner and they are indeed package even if they didnt come to onemac so it does make sense in a way.

Approved.

PR #1251, opened 2023-05-18
Oy2 22277 - split TEs into two types for 1915c and 1915b

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-22277

### Details

Temporary extensions are divided into types based on selection of 1915c. or 1915b on the sumbission form. They still share the same UI form and therefore same api submission endpoint. At point of submission the api will determine what configuration type to submit based on the selected temporaryExtensionType.

This change also is migrating existing temporary extensions to their proper type. This is done in two parts. One from a list of TEs that have been manually mapped (those that existed before package release and therefore do not have a temporaryExtensionType attribute) and the other for all TEs submitted after package release and the type is present and therefore can be used to udpate componentType to the proper value.

### Changes

- I think the test coverage diff is so low because of the migration script. Do we care?

### Test Plan

1. login as state submitter
2. submit new temporary extension using type 1915(b)
3. verify new package on dashboard shows appropriate type label
4. repeat for 1915(c)


services/common/workflow.js
```diff
@@ -14,6 +14,8 @@ export const ONEMAC_TYPE = {
   WAIVER_RENEWAL_WITHDRAW: "waiverrenewalwithdraw",
   WAIVER_RAI: "waiverrai",
   WAIVER_EXTENSION: "waiverextension",
+  WAIVER_EXTENSION_B: "waiverextensionb",
+  WAIVER_EXTENSION_C: "waiverextensionc",
```
kristin-at-theta, line 18: I like that you kept the plain "waiverextension" type as well, since we still have those leftover packages without a b or c...

kristin-at-theta (approved):
I do think it is weird that you can use the Action dots to request an extension for a package on the dashboard and it pre fills the ID, but not the type - I was able to create a 1915c extension on a 1915b waiver.  That detail is not covered in the AC, in fact, looks like the AC is only talking about the dashboard type - though I did check and the details pages are also updated (which I think is right)

Code looks good, tested the forms and the migrate and the types look correct to me!

Heh, we really should add migrate.js to the codeclimate ignore list.... :) 

Approved!

PR #1252, opened 2023-05-18
Lambda Refactor: Rebuild packages

kristin-at-theta:
Story: Lambda Refactor

### Details
the rebuildPackages lambda is supposed to go through all the SEA Tool topics and update them such that the packages get rebuilt.  However, it quickly reaches a write throttling threshold, so is unreliable.

### Changes
- created a scheduled lambda that looks for "jobs" and runs them
- rebuild packages now adds jobs to the queue

### Implementation Notes
- I guessed with many of the variables, so might need tweaking, either to speed up the rebuild, or if we hit the threshold again.

### Test Plan
1. Login to AWS Console
2. Go to the lambda console for one-stream-rebuild-packages-rebuildPackages
3. Navigate to Test tab
4. add "reason": "testing" (can be any reason, really) to the test event
5. run the test
6. Verify that the Process items have been added to the one table
7. Verify that processPackages picks up the Process items, runs, then updates the Process items with the new start keys
8. Try to watch for throttling issues or if anything isn't working (I know, sorry!)

### DIFF
--- a/services/one-stream/processPackages.js
+++ b/services/one-stream/processPackages.js
@@ -1,0 +1,92 @@
+import AWS from "aws-sdk";
+
+import { dynamoConfig } from "cmscommonlib";
+
+const dynamoDb = new AWS.DynamoDB.DocumentClient(dynamoConfig);
+
+const oneMacTableName = process.env.IS_OFFLINE
+  ? process.env.localTableName
+  : process.env.oneMacTableName;
+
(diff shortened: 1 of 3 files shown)


bflynn-cms (approved):
Discussed with Kristin about reason for this approach which is limited by how dynamodb ondemand throughput functions.

Approved.

PR #1253, opened 2023-05-19
OY2 23578: Update CMS Package Details pages with CPOC + SRT fields from SEA Tool

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23578
Story: https://qmacbis.atlassian.net/browse/OY2-23647

### Details
CPOC (CMS Point Of Contact) is a field from SEA Tool that OneMAC would like to display on the details pages and as a column on the dashboard.  This is for all users, not just CMS (see [23647](https://qmacbis.atlassian.net/browse/OY2-23647))

SRT is the Subject Review Team - a list of contacts within CMS who are reviewing the package.  This list is only shown to CMS users and has a new format, a expandable list of names.

### Changes
- Updated the one-seed.json to include the CPOC and SRT in the package seed items.
- include the CPOC and the SRT team in the package item
- pass the CPOC in both user responses
- pass the SRT list in CMS user responses
- add the CPOC as a default package detail into the UI
- add the SRT as a new component utilizing the detail design

### Implementation Notes
Because the SRT is the final detail in the Detail Section, we can leave the current more-generic detail processor as-is and add the SRT as a state-aware component with the same design to the end of the details.  If we end up needing to alter this order... we will need to refactor how the detail Reviews are constructed and allow them to contain components with state.

### Test Plan
1. Login as a state submitting user
2. Navigate to dashboard
3. Verify that CPOC Name is a column that can be shown/hidden with default shown
4. Verify that clicking into the package shows the CPOC as a detail per the Figma and story AC
5. Verify that the Review Team (SRT) is NOT shown to the state user.
6. Logout/Login as a CMS user
7. Navigate to dashboard
8. Verify that CPOC Name is a column that can be shown/hidden with default shown
9. Verify that clicking into the package shows the CPOC as a detail per the Figma and story AC
10. Verify that the Review Team (SRT) IS shown to the CMS user.
11. Verify that if the Review Team contains more than 3 names, it is shrunk with an option to expand/show all
12. Verify that show/hide the Review Team works as expected.


services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -116,6 +125,34 @@ export const DetailSection = ({
...
+                    <>
+                      {reviewer}
+                      <br />
+                    </>
+                  ))
+                : detail.reviewTeam.slice(0, 3).map((reviewer) => (
+                    <>
+                      {reviewer}
+                      <br />
+                    </>
```
bflynn-cms, line 143: Im not a fan of the repeated block to show the reviewer........but i guess the only way around that would be to always do a slice with the second parameter being a dynamic var that is toggled between list size and NUM_REVIEWERS_TO_SHOW when the show/hide is toggled along with the showReviewTeam boolean. As is I think the slice should use the declared var and not a hard coded 3.

services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -116,6 +125,34 @@ export const DetailSection = ({
...
+              {detail.reviewTeam.length > NUM_REVIEWERS_TO_SHOW && (
+                <Button
+                  className="review-team"
+                  aria-expanded="false"
+                  onClick={toggleReviewTeam}
+                >
+                  {showReviewTeam ? "View Less Names" : "View All Names"}
+                </Button>
+              )}
+            </Review>
```
bflynn-cms, line 154: I like the fore thought about only showing the button if the size is greater...makes perfect sense and might be something I wouldnt have thought about until I ran into the condition live in the app!

services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -22,6 +30,7 @@ export const DetailSection = ({
   setAlertCode: (code: string) => void;
 }) => {
   const { userProfile } = useAppContext() ?? {};
+  const [showReviewTeam, toggleReviewTeam] = useToggle(false);
```
bflynn-cms, line 33: I learned something new! I didnt know about useToggle

bflynn-cms (changes requested):
Looks great! Everything tests out well!

Made a few comments along the way but only one real request to use the NUM_REVIEWERS_TO_SHOW var in the slice for displaying reviewers instead of a hard coded value. 

Also made a comment in the same location of the slice on a suggestion of code redundancy, but that is merely a suggestion that may or may not be a good idea.

services/ui-src/src/page/section/DetailSection.tsx
kristin-at-theta: you can thank George for useToggle :) it is a custom hook he added back in the day for the Phone Number editing

services/ui-src/src/page/section/DetailSection.tsx
kristin-at-theta: I am also not a fan of the repeated block - but I was having SO MUCH trouble getting the logic right when I was "being clever" that I decided getting it right was more important than pretty :)   Though I absolutely agree about the slice 3 should use the constant - my only defense is that I did the constant *after* doing the slice and simply forgot... :O

services/ui-src/src/page/section/DetailSection.tsx
kristin-at-theta: .... which is what happened :)  I was like "what's this?!" during my own testing

services/ui-src/src/page/section/DetailSection.tsx
kristin-at-theta: Apparently I channeled George this morning... sorry for making it a bigger re-review... but figured an internal component would isolate out the toggle and also remove the duplication... of course, I made a bunch of silly mistakes about which length was which in the process... but I *think* it is right now!

services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -9,6 +14,47 @@ import FileList from "../../components/FileList";
...
+}) => {
+  const [showExpanded, toggleExpanded] = useToggle(false);
+
+  if (!list) return <></>;
+
+  return (
+    <Review key="expanded-list" heading={heading}>
+      {list.length === 0
+        ? "-- --"
+        : list.slice(0, showExpanded ? list.length : numToShow).map((line) => (
```
bflynn-cms, line 40: love this!

bflynn-cms (approved):
I love the new changes to componetize the list....makes the actual jsx being returned by the detail component much easier to view. Also like the refactor to avoid duplication of output for each item!

Approved.

PR #1254, opened 2023-05-22
clear table state if not authenticated (upon redirect to login page)

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23899

### Details

Currently table filters (state) is cleared upon a timeout and an intentional logout action; however it was not being cleared upon a token timeout if the user did not have the page loaded at the time. This solution will cause the state to be cleared on any route change (such as on app load if user uses a direct url navigation or reopens the app from start) by weaving into the logic already built into our routing for unauthenticated users.

### Changes

- updated routing logic for when an unauthenticated user (including expired token) loads a route then the table state will be cleared

### Test Plan

1. Login as any user
2. Set table filters
3. close tab and wait for token to expire
4. Open new tab and you should be redirected to login page
5. login again and verify table state is reset


kristin-at-theta (approved):
Took forever to finally get where the IdleTimer actually logged out without putting the application in some very weird states... one, the page was entirely blank:
Screenshot 2023-05-22 at 12 35 09 PM

Then... it "logged me out" but I was still able to access everything, but the React Components had errors....

Finally, it really did log me out and it does look like the table settings got reset! 

I guess there is a possibility there is some timing issues?  is anything in clearTableStateStorageKeys asynchronous??  I remember we had a heck of a time with logout back in the day... (and still I don't think we do it right)

Anyway, I finally got it to kinda do what I thought... and I agree with the code logic... so... Approved!

Though.... please fix the unit test coverage, if possible... looks like the other condition is tested, so maybe can add a test with the other condition?

bflynn-cms:
> 

@kristin-at-theta I updated the location in routes where I was doing the reset because I realized some authenticated routes also used the signupGuardRouteListRenderer.....so i moved the logic up one level to be triggered on the parent of both renderers. Just heads up in case you wanted to double check the changes.

Also by doing so the tests are now 100% coverage on changed lines.

kevinhaube (approved):
Boop! Ez win. I didn't get the same bug Kristin did, so I'm assuming that's those changes you mentioned. 👍🏼

PR #1260, opened 2023-05-31
Oy2 24002

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-24002

### Details

Rework how email addresses operate such that any function can override the ccEmail property in serverless config without having to add a new structure to form config

### Changes

- Made default form config look at process.env.ccEmail rather than only override on certain forms
- Made chip special emails for reviewer and cc use standard reviewerEmail and ccEmail rather than different env var names
- updated form configs to follow the above approach
- added new ssm parameter store properties for the 3 new email groups
- updated serverless config to set proper email groups per method for ccEmail

### Implementation Notes

- As these changes rewire how env vars are set along with adding new ssm parameters; the values will need to be manually set after release to prod

### Test Plan

1. Verify each form type still submits
2. Verify setting reviewerEmail and ccEmail on each of the affected form types sends to proper email address

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -26,6 +26,9 @@ custom:
   reviewerEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_spa_form_cms_email, ssm:/configuration/default/email/cms_spa_form_cms_email, file(resources/ssm-params.yml):Resources.CmsSpaFormCmsEmail.Properties.Value}
   reviewerCHIPEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_spa_form_chip_email, ssm:/configuration/default/email/cms_spa_form_chip_email, file(resources/ssm-params.yml):Resources.CmsSpaFormChipEmail.Properties.Value}
   chipCcEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_chip_form_cc_email, ssm:/configuration/default/email/cms_chip_form_cc_email, file(resources/ssm-params.yml):Resources.CmsSpaFormChipCcEmail.Properties.Value}
+  dmcoEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_dmco_email, file(resources/ssm-params.yml):Resources.CmsDmcoEmail.Properties.Value}
+  dhcbsooEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_dhcbsoo_email, file(resources/ssm-params.yml):Resources.CmsDhcbsooEmail.Properties.Value}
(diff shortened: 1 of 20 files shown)


services/app-api/serverless.yml
```diff
@@ -26,6 +26,9 @@ custom:
   reviewerEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_spa_form_cms_email, ssm:/configuration/default/email/cms_spa_form_cms_email, file(resources/ssm-params.yml):Resources.CmsSpaFormCmsEmail.Properties.Value}
   reviewerCHIPEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_spa_form_chip_email, ssm:/configuration/default/email/cms_spa_form_chip_email, file(resources/ssm-params.yml):Resources.CmsSpaFormChipEmail.Properties.Value}
   chipCcEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_chip_form_cc_email, ssm:/configuration/default/email/cms_chip_form_cc_email, file(resources/ssm-params.yml):Resources.CmsSpaFormChipCcEmail.Properties.Value}
+  dmcoEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_dmco_email, file(resources/ssm-params.yml):Resources.CmsDmcoEmail.Properties.Value}
+  dhcbsooEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_dhcbsoo_email, file(resources/ssm-params.yml):Resources.CmsDhcbsooEmail.Properties.Value}
+  dpoEmail: ${ssm:/configuration/${self:custom.stage}/email/cms_dpo_email, file(resources/ssm-params.yml):Resources.CmsDpoEmail.Properties.Value}
```
kristin-at-theta, line 31: Looks like this default fall-through is how we keep redeployments from overwriting values already in the environment. Nice!

kristin-at-theta (approved):
Definitely appreciate the consolidation of the CHIP/everyone else email variables - after we separated the handlers, didn't make much sense to carry two sets of those... nice cleanup :) 

Rather than waiting until after deployment to set the emails (something that sounds like it will be forgotten...) could the variables be preset to the right values in the production account so deploying to production automagically gets the correct values?

I'm not keen on real email addresses being published in a public repo - maybe use a monitored dummy address that would trigger someone thinking "huh, we've got an issue" -  we don't currently have a team email address for stuff like that, but that's what I've used in the past.  Something to think about.

Tested that updating the email address does send the withdraw email to that address... which is what the code re-organized, so I believe the PR is complete.  I feel like the story AC is impossible to verify until this gets all the way to production.  So, I'm not saying this PR has effectively satisfied the AC of the underlying story, but the modifications are good, well-constructed, and pave the path for the real AC.

Approved!

PR #1261, opened 2023-05-31
Two Column Details

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-24007

### Details

This PR turns the single-column of detail items on a details page into a two-column view.

Before:
Screen Shot 2023-05-31 at 3 33 03 PM

After:
Screen Shot 2023-05-31 at 3 32 30 PM

### Changes

- Use of `Grid` from Material-UI to handle the CSS and logic for sizing and scaling
- Use of `Review` remains from the original implementation -- is wrapped in grid structure

### Implementation Notes

- Due to how we configure and display our details sections -- namely, as arrays of items to be displayed rather than an object with named attributes -- it's an expensive effort to process the array and display items in specific portions of the grid as requested in the ticket's AC. This affects the ability to accurately place the divider, and accurately showcase items in the columns defined in the AC.
- The left edge of the right-hand card does not align with the grid due to a difference in Material-UI's breakpoint/scaling system and our own. This can be considered a visual bug and we can make at ticket to address this by [customizing MUI's idea of breakpoints and spacing.](https://v4.mui.com/customization/breakpoints/)

### Test Plan

1. Navigate to a package's details page
2. Ensure the view is as described by the AC (sans the exception in Dev Notes)
3. Ensure this is consistent across various package types' detail pages


services/ui-src/src/libs/detailLib.ts
```diff
@@ -110,19 +118,31 @@ export const defaultPackageOverviewNavItems: DetailNavItem[] = [
...
-  typeDefault,
+  submissionIdDefault,
+  blankBox, // empty space
   territoryDefault,
+  typeDefault,
   submissionDateDefault,
-  latestRaiResponseDateDefault,
   proposedEffectiveDateDefault,
+  latestRaiResponseDateDefault,
+  blankBox, // empty space
```
bflynn-cms, line 128: Could we use a nested array here as the type definition of a "SectionItems"? Whereby the first array is the first column and the second array is the second column. That way we could define something like:
export const defaultDetailSectionItems = 
[
  [field1, field2, field3, field4],
  [blankBox, field 5]
 ]
 
 i suppose the downside is we are building layout into the config when thats the job of the page......but id argue we already made that choice to genericize it by enforcing layout with the order of the detail items in the array. This setup would lead nicely into how you are already using a double nested array in the detail component and would also get rid of the moving column issue for unrendered data in column A.

bflynn-cms (changes requested):
Made a comment on possible suggestion for simplification.
Other than that code looks good!

As for testing I did have a question regarding Subject and Description, I am seeing these fields for State user as well, whereby ACs show they should only be for CMS users? I double checked develop branch as well and only see those fields when logged in as a CMS user.

Also AC indicates they wanted a lined border between standard fields and the CPOC and SRT fields? Maybe you talked them out of that 90s style decision but AC didnt get updated?

services/ui-src/src/libs/detailLib.ts
kevinhaube: Hey! There's actually going to be another unit of work (see [Erika's comment](https://qmacbis.atlassian.net/browse/OY2-24007?focusedCommentId=121141) on JIRA) that will put the layout strictly in the hands of the component. After talking with Erika, we determined the AC for this work would be to simply take what we have now, and make it two-columns and "as close as we can to the Figma" -- in this case, the divider, Description being a full-width-single-column, and some empty spaces were cut until we can layout the page based on an object with defined attributes and not the array we have now.

As for the permissions, I don't _think_ any of that code got removed...that's odd, and I'm looking into it

kevinhaube: Woops -- those fields just didn't have a role restriction applied to them is all. Tho...that's curious, because if they _don't_ show now for State users in Dev/Val and they didn't have the `rolePrivilate` attribute, how were they hidden? lol

bflynn-cms: sounds good! I wish we could get ACs updated by business to reflect discussions in the comments..... and yeah i couldnt see why your changes would have changed the permissions on those fields either.
Thanks!

services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -76,16 +77,12 @@ const TwoColumnDetails = ({
...
-          <Grid xs={12} sm={5} item className="detail-grid-item">
-            <DetailItem item={secondColumn} />
-          </Grid>
-        )}
+        <Grid xs={12} sm={5} item className="detail-grid-item">
+          {hasPermission(firstColumn) && <DetailItem item={firstColumn} />}
+        </Grid>
+        <Grid xs={12} sm={5} item className="detail-grid-item">
+          {hasPermission(secondColumn) && <DetailItem item={secondColumn} />}
+        </Grid>
```
bflynn-cms, line 85: i like this change to always declare the grid element and only render inside if you have permission to view the element! Nice way to keep the defined position!

bflynn-cms (approved):

kristin-at-theta:
For Information:
The showing/hiding of fields is/was being managed by the back end (getDetail). A flag, canSeeSubjectAndDescription, is on the user role object and if false, the attributes are deleted from the item before the data is sent.  That way, you can't read the details by looking in the response.

PR #1262, opened 2023-06-01
align alert bar dimensions to be same as header

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23918

### Details

Update format of alert bar to have max size same as header

### Changes

- Css updates to alert-bar and ds-c-alert classes

### Test Plan

1. login as state submitter
2. submit any new package
3. verify alert bar sizing

### DIFF
--- a/services/ui-src/src/index.scss
+++ b/services/ui-src/src/index.scss
@@ -346,10 +346,17 @@ ul {
 
 .alert-bar:not(:empty) {
   padding-top: 2rem;
-  padding-left: 14rem;
-  padding-right: 14rem;
+  padding-left: 7.2rem;
+  padding-right: 7.2rem;
   padding-bottom: 2rem;
 }
+
+.ds-c-alert {
+  @extend .ds-u-margin-left--auto;
+  @extend .ds-u-margin-right--auto;
+  @include max-width-ultra-wide;
+}
+
 .alert-bar button {
   position: absolute;
   background: inherit;


kristin-at-theta (approved):
Yeah, looks fine to me as presented in the AC (though I would not say the margins are container-specific, since (if I remember correctly) the page is stacked containers sharing the same base style).

Of course, when you pay attention to design, you start to see the design :) Not sure how the Action column got put "outside the container," but I don't think that is correct...
Screenshot 2023-06-01 at 8 15 01 AM

Also, I believe there is another story this sprint to change the padding/margins of the headers and pagetitlebar, please also get the alert bar at the same time :) 
Screenshot 2023-06-01 at 8 16 36 AM

Approved!

bflynn-cms:
> 

oh forgot to mention, from your first screenshot we do allow the table to go outside the normal bounds and I dont think we want to align the filters or alert with the full table? I do like how we had it before though where the table expanded out from center and would spill evenly both left and right......though I guess that is probably less desirable with smaller screens as it would start to spill off screen left which would be no bueno. At any rate the current design has the header/alert/filter set to a max width of 1024px which I think is a good choice otherwise it starts to become too much whitespace in the middle.

bflynn-cms:
Closing this as it is OBE and ultimately the banner fix was rolled into https://qmacbis.atlassian.net/browse/OY2-23528

PR #1264, opened 2023-06-02
add FAQ section and link from triage landing pages

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-24094

### Details

Add new FAQ section and link to section from triage landing pages

### Changes

- Added new FAQ section
- Added new Route for FAQ section
- Added link to FAQ Route on triage page

### Test Plan

1. Login as state submitter
2. Go thru triage to land on each external system landing page
3. verify link navigates to FAQ with section expanded
4. verify clicking link to document in FAQ downloads expected file


kristin-at-theta (changes requested):
(change requested) pdf does not download in my Google Chrome, I get:
Screenshot 2023-06-02 at 8 15 24 AM

(requested "look into" and change, if possible) If I scroll down in the "new" window, then return to the landing page and use the FAQ link for the crosswalk, the question is scrolled off the screen, if there is no "new" window available, the question is shown at the top:
Screenshot 2023-06-02 at 8 19 44 AM

Must've missed this in earlier FAQ stories.... ROUTES was deprecated with Submission View - we should be using ONEMAC_ROUTES - at least, that was my understanding of how it would work... if we are not using only one ROUTES object, should document when to use one versus the other. (this is not a requested change)

kristin-at-theta (approved):

kevinhaube (approved):
Woo! Looks good to me :)

PR #1265, opened 2023-06-04
oy2-235228 - fix margins to be same for all content and headers under 1024px wide

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23528

### Details

Bandaids and duct tape to make css behave without restructuring......i think i used bubblegum too

### Changes

- CSS changes only

### Test Plan

1. Login as any user
2. Navigate to package dashboard and observe behavior of all page elements margins for various screen dimensions
3. Navigate to all other page types to verify header/content behave as expected (package submission, package details, user profile, etc)

### DIFF
--- a/services/ui-src/src/index.scss
+++ b/services/ui-src/src/index.scss
@@ -180,8 +180,6 @@ p.field-hint {
 }
 .app-bar {
   @extend .ds-l-row;
-  padding-left: 7.2rem;
-  padding-right: 7.2rem;
   @extend .ds-u-padding-top--1;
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
Margins react the way I expect!

I do think it is weird that there is a size that requires tweaking, but ... eh... design can be weird! :) 

Actions column outside the page is a known issue, and is still presenting here.

Approved!

kevinhaube (approved):
Scales correctly, and CSS looks good to me 👍🏼

PR #1274, opened 2023-06-09
Update target for all FAQ links

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-24192

### Details

This PR unifies all FAQ links so they open in `target="tab-faq"` by way of a new exported variable in `routes.js`, `FAQ_TARGET`. 

### Changes

- New constant in `common/routes.js`
- Change value for `target` on FAQ links

### Implementation Notes

N/A

### Test Plan

1. Ensure FAQ nav link opens in new tab
2. LEAVE THAT TAB OPEN! 😄 
3. Navigate back to the original tab.
4. Ensure FAQ link in footer opens in the existing FAQ tab.
5. Repeat steps 3-4 for FAQ links from a Form (these are procedurally created, so one form should be fine for testing)
6. Repeat steps 3-4 for FAQ inks from triage landing pages (i.e. the pages that link to MMDL and MacPRO)

### DIFF
--- a/services/common/index.d.ts
+++ b/services/common/index.d.ts
@@ -1,6 +1,11 @@
 export const RESPONSE_CODE: Record<string, string>;
 export const FORM_SUCCESS_RESPONSE_CODES: string[];
-export { ROUTES, ONEMAC_ROUTES, TYPE_TO_DETAIL_ROUTE } from "./routes.js";
+export {
+  ROUTES,
+  ONEMAC_ROUTES,
+  TYPE_TO_DETAIL_ROUTE,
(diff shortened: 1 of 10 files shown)


kristin-at-theta (approved):
Nice to have a constant value and I agree with choice to put value in routes.

Tests correctly, did a search on "targets" and "blanks" and looks like the right ones were changed and the out-of-OneMAC travel is sent to blank (which is correct, UNLESS we end up wanting to chase the analytics across applications or something, since a named window gives us more options there, but I digress)

Approved!

PR #1281, opened 2023-06-14
OY2 24262: Splitting packages can mean NOSOs that pull attachments from an OSO

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-24262

### Details
Splitting packages can mean that the attachments are relevant to both packages, so use the NOSO lambda with an extra ability to grab the attachments from a different package.

### Changes
- if "copyAttachmentsFrom" is designated in the NOSO event, copy the attachment information from the given package.
- updated the createOneMACPackage.json to include an example usage of this variable

### Test Plan
1. Login to OneMAC and find a package you would like to "split"
2. Login to AWS Console and navigate to Lambda > Functions > admin-oy2-24262-createOneMacPackage lambda
3. copy the sample JSON into the test event area
4. update the sample to reflect the package details for Instruction 1
5. run the test
6. Refresh the dashboard
7. Verify the new package is on the dashboard with the details submitted and contains the attachments from the starting package.

### DIFF
--- a/services/admin/handlers/createOneMacPackage.js
+++ b/services/admin/handlers/createOneMacPackage.js
@@ -64,7 +64,7 @@ export const createOneMacPackage = async (item) => {
     };
     try {
       const result = await dynamoDb.get(getParams).promise();
-      putParams.Item.attachments = { ...result.Item.attachments };
+      putParams.Item.attachments = [...result.Item.attachments];
     } catch (e) {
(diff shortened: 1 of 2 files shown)


bflynn-cms (approved):
Looks good and my medicaid spa and initial waiver test worked as expected and copied over the attachments from the specified package.
Approved!

PR #1286, opened 2023-06-15
OY2 24058: BUG - When the ID isn't found, OneMac shows an empty details page

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-24058

### Details
if you change the package id in the url of the details page, it showed a weird, empty, details page.  We would rather it show a "page not found"

### Changes
- added the "NotFound" component to the details page
- restructured the page so that if the componentId is not in the response for fetch details, display "NotFound"

### Implementation Notes
- Added the component instead of manipulating the routes or history because I thought it would be better to maintain the URL - but that's just personal preference.

### Test Plan
1. Login and navigate to Dashboard
2. Select a package from the table and open its details
3. Click into the URL and change the ID to something not in the database
4. Verify "NotFound" is shown and looks like other "Page not found" pages
5. change the URL to a valid package
6. Verify detail page is correct again.


bflynn-cms (approved):
Code looks good and behaves as expected.
Nice cleanup on the error redirection!
Approved!

PR #1287, opened 2023-06-20
Oy2 24437 - generate signed urls for withdrawal attachments on detail view

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-24437

### Details

Generate signed urls for attachments on withdrawal requests so that they can be downloaded.

### Changes

- Add logic to generate presigned urls for withdrawalRequests object
- Moved all generation of all signed urls (package, rai response, withdrawal requests) to be within single method 

### Implementation Notes

- I did try a fancy pants recursive function that would look at an object for an attachments array, generate urls, and then call itself for each remaining property. This could have auto covered any future additions of attachments as sub objects however turned out to be too fancy pants and as recursion usually is was not as easy to follow or easily root out side issues. I figured that was a sign to do it simply and declaratively with known objects. :)

### Test Plan

1. Login as state submitter
2. Create a package with attachments and set its status to Under Review or locate an existing package
3. Perform withdraw action on the package and include an attachment
4. Open details page and verify the main package attachments and the withdrawal attachment can be downloaded.
5. Also try an rai response just to be sure.

### DIFF
--- a/services/app-api/getDetail.js
+++ b/services/app-api/getDetail.js
@@ -8,26 +8,36 @@ import { validateUserReadOnly } from "./utils/validateUser";
 
 const s3 = new AWS.S3();
 
+async function generateSignedUrl(item) {
+  const attachmentURLs = await Promise.all(
+    item.attachments.map(({ url }) =>
+      s3.getSignedUrlPromise("getObject", {
+        Bucket: process.env.attachmentsBucket,
+        Key: decodeURIComponent(url.split("amazonaws.com/")[1]),
+        Expires: 3600,
+      })
+    )
+  );
+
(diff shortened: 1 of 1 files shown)


kristin-at-theta (approved):
Ok, so your attempt at recursion explains the consolidation into the attachmentsURL function - which is fine!  I probably would've been tempted to try the same thing :) maybe...

Anyway, code looks correct, it deals with the current groups of attachments, and it is clear where to add new submission types with attachments (like this Formal RAI Response Withdrawal Request - say that 10 times fast!)

Yes, I also tested that the other, older attachments still work - though was reminded that the seed data does NOT have valid attachments.

Approved!

PR #1289, opened 2023-06-21
oy2-24098 - update migrate with new set of TE migration mappings

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-24098

### Details

Add more TE type mappings to migrate TE to subtype

### Changes

- removed prior migration function to migrate based on user choice (all should be mapped and no new ones can be created)
- updated map collection with set provided in ACs

### Test Plan

1. ??? Not really testable as these are production specific packages; suppose one could create a package with same name and edit data to give it the old generic type but that would just be retesting the migration we already did on first set in prod.


kristin-at-theta (approved):
The logic around why not to retest the mapping is sound - and looking at the code, there was only cleanup of extra no-longer-needed functionality and none of the code was changed for the mapping.  thanks for the cleanup!

Verified that the list in temporaryExtensions is the same ID and extensionType as in the story.

Approved!

PR #1290, opened 2023-06-23
OY2 24270: Admin Change History

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-24270

### Details
Add the Admin Change History to the package details.

### Changes
- admin lambdas, softDeleteComponent, createOneMacPackage, batchCreateOMP, and updatePackageId now add an adminChanges entry
- build packages pulls all the adminChanges into one array and sorts reverse-chrono
- ui updated to show the Admin Change History as a section similar to the RAI Responses
- seed data includes a couple adminChanges for testing/viewing
- added adminChanges to test data for unit test coverage.

### Implementation Notes
- when hand-editing data, make sure to include this record of the changes made in the event that got edited.
- editing packages directly is a bad idea, as they get overwritten by the supporting events.  figure out which event should change to push the package change.

### Test Plan
1. Login to OneMAC as a user that can view packages
2. Verify that the package details of admin-edited packages contain the Admin Change History section
3. Verify that the package details of packages NOT edited by the admin DO NOT contain the Admin Change History section
4. Login to the AWS console
5. Change some packages in the feature branch using the admin lambdas.
6. Verify the packages have the correct Admin Change History details in their details.

NOTE: the story specifies EST for the date/time, but that is not how any of our other date/times are done.  Browsers automatically covert to local time.  The date/time was left to match the other uses.


bflynn-cms (changes requested):
Style, format, positioning, opening of first history item, hiding if no items all look good.

I did encounter an issue when doing an admin changePackageId funciton though in that I am seeing duplicate records for each action. I was able to duplicate this duplication a few times. I did not yet test any of the other admin actions.

kristin-at-theta:
the duplicate entries are somewhat purposeful... as we can only make admin changes to events, so changing the PackageID is really changing all the IDs in each event... AND we need to keep it that way because sometimes it is the RAI response ID that gets edited to MATCH it to the original submission... 
In any case, the duplication does Look Weird so I updated the package builder to remove duplicates based on the timestamp... which means the "Admin Changes" are not exactly correct for all the elements of the package, but does roll up the changes and does not confuse the user with duplicates - maybe hopefully??

bflynn-cms (approved):

bflynn-cms:
> the duplicate entries are somewhat purposeful... as we can only make admin changes to events, so changing the PackageID is really changing all the IDs in each event... AND we need to keep it that way because sometimes it is the RAI response ID that gets edited to MATCH it to the original submission... In any case, the duplication does Look Weird so I updated the package builder to remove duplicates based on the timestamp... which means the "Admin Changes" are not exactly correct for all the elements of the package, but does roll up the changes and does not confuse the user with duplicates - maybe hopefully??

yeah I see what you mean, I think how you hide the duplicates makes sense. Thanks!

PR #1292, opened 2023-06-23
Add a source service to stream onemac data to kafka

mdial89f:
Story: https://qmacbis.atlassian.net/browse/OY2-24530
Endpoint: https://d26w3prdxhme46.cloudfront.net/

### Details

This changeset adds a new service named 'source'.  It is comprised of a lambda set to listen to onemac's onetable stream, and send changes to a kafka topic.

### Changes

- source service added
- sourceDynamoToMsk function was created to filter out seatool data and stream what's left to a kafka topic.
- the kafka topic is named aws.onemac.migration.cdc
- the source service will only deploy for develop, master, and production branches.  It supports deploying on ephemerals, but I figured it's unneeded.

### Implementation Notes

- A one time migration of existing data not represented in the stream needs to occur upon promotion.  On that note, @mdial89f should be notified before merge, to handle this step.

### Test Plan

The events are currently streaming as expected.  No further plan developed.


benjaminpaige (approved):
Wonderful addition

services/source/handlers/cleanupKafka.js
Dark-Knight-1313, line 1: Do we want to add an actual timeout value for the deleteTopics or are we good with the default API timeout?

Dark-Knight-1313, line 1: Just note: We are taking the second table stream and Dynamo only allows two stream subscriptions so if we ever need anything else to use table streams we will have to reconstruct using a fan out pattern.

Dark-Knight-1313 (approved):

deploy.sh
Dark-Knight-1313, line 1: Unit tests? Cypress automated tests?

PR #1294, opened 2023-06-26
Add Character Counter w/ Aria for Screen Reader

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23575

### Details

This PR adds a 4000-word character counter to the Additional Information section of OneMAC forms. It's built to debounce the read-back using the `aria-busy` tag and React state logic. The debounce is set to 500ms per the AC. Using `aria-live` off, immediate updates to the additional information will not be read, and updates to the character count will only happen if nothing else is being read out (i.e. "polite" setting). 

### Changes

- Adds Character Counter
- Adds aria tags to character counter for proper readout

### Implementation Notes

- I'm not sure 500ms is long enough of a debounce, but the AC dictates all.

### Test Plan

1. Turn on a screen reading tool like MacOS's VoiceOver
2. Navigate to a form page with an additional info box
3. Assert the character counter is present
4. Assert the character counter reads your count after you're done typing

### DIFF
--- a/services/ui-src/src/page/OneMACForm.tsx
+++ b/services/ui-src/src/page/OneMACForm.tsx
@@ -63,6 +63,15 @@ const OneMACForm: React.FC<{ formConfig: OneMACFormConfig }> = ({
   const { activeTerritories, confirmAction } = useAppContext() ?? {};
   const location = useLocation<FormLocationState>();
 
+  // Boolean controlling aria-busy for Additional Info section
+  const [isTyping, setIsTyping] = useState<boolean>();
+  // Effect to timeout and swap isTyping back to false
+  useEffect(() => {
+    const timeTilInactive = setTimeout(() => setIsTyping(false), 500);
+    return () => {
+      clearTimeout(timeTilInactive);
+    };
+  }, [isTyping]);
   //Reference to the File Uploader.
   const uploader = useRef<FileUploader>(null);
   // True when the required attachments have been selected.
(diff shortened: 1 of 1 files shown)


bflynn-cms (commented):
Im not sure I know how to accurately review this change. As mentioned 500ms is way too short of a span and I think we should consider changing that.
Also I didnt see any consistency on when the screen reader would read the total character count....it seemed totally random and often in between keystrokes, however; using the usdws example it was pretty consistent. I feel like there is something off there.

kristin-at-theta:
I tried Google Screen Reader, the plain mac accessibility reading, and the mac VoiceOver and at no time did it speak about the number of characters either typed or left.

kevinhaube:
@kristin-at-theta Yeah, I tweaked the aria stuff but it hasn't redeployed because of an issue with the Deploy action :/

kevinhaube:
@kristin-at-theta @bflynn-cms Looks like the update fixed the deploy -- should be good to test now!

services/ui-src/src/page/OneMACForm.tsx
```diff
@@ -523,10 +532,20 @@ const OneMACForm: React.FC<{ formConfig: OneMACFormConfig }> = ({
...
+            }}
             value={oneMacFormData.additionalInformation}
             maxLength={config.MAX_ADDITIONAL_INFO_LENGTH}
+            aria-describedby="character-count"
+            aria-live="off"
+            aria-multiline={true}
+            aria-busy={isTyping}
           ></TextField>
+          <span tabIndex={0} aria-label="character-count" aria-live="polite">
+            {oneMacFormData.additionalInformation.length}/4000
```
kristin-at-theta, line 547: Would changing this to {4000-oneMacFormData.additionalInformation.length} remaining of 4000 be better?  I know the Figmas have the 0/4000, but AC does specify the remaining, not the used...

kristin-at-theta (changes requested):
So... I think it is doing what is supposed to in terms of telling you at an appropriate time and not after every character....

Change Requested... the AC in the story says "how many chars are left" and what VoiceOver and Google Screen Reader are reading is "how many characters have I typed"

kevinhaube:
@kristin-at-theta Great catch! I fixed it and also managed to get screen readers to recognize the text "characters remaining" so it doesn't just shout a number out lol

Updates pushing now :) 

Screen Shot 2023-07-06 at 2 01 55 PM

kristin-at-theta (approved):
So much better!  In fact, it was kinda even useable with the Google Screen reader - it read my letters, then if I stopped, would read what I wrote and then tell me the characters remaining!  I feel like that would be much more useful (though I admit, it is hard to put myself in their shoes on this)

Approved!

PR #1296, opened 2023-06-30
OY2 24303: TE Request should pre-fill TE Type

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-24303

### Details
Since the TE type depends on the Waiver package being extended... makes sense that when starting a Temporary Extension Request from an existing package, we pre fill the Type. Type is associated with the Waiver Authority, but is at a higher level, so 1915(b) and 1915(b)(4) are both 1915(b)s for the purpose of Extensions.

### Changes
- added temporaryExtensionType as a var passed in the FormLocationState
- getMyPackages and getDetails now return temporaryExtensionType as part of the package item
- modified TemporaryExtensionTypeInput to accept a pre fill and "disable" the component in the same way we do the ComponentIds

### Implementation Notes
- Did use the fact that 1915(b) and 1915(b)(4) both start with 1915(b) to set the value for Temp Extension Type
- may want to move this to package build time? 

### Test Plan
1. Login as a submitting user
2. Navigate to Waivers tab on Dashboard
3. Find an Initial Waiver or Waiver Renewal in "Approved" status
4. click the Action dots, verify the type is prefilled
5. Verify can submit the Temporary Extension
6. Return to Dashboard
7. Click into the package details of an appropriate package
8. Click the "Request Temporary Extension" link in Package Actions
9. Verify the type is prefilled
10. Verify you can create the extension
11. use the New Submission button 
12. Triage to Temporary Extension Form
13. Verify the blank version of form still works as expected.


kevinhaube (approved):
Code looks good to me, and it functions in the branch!

services/ui-src/src/page/temporary-extension/TemporaryExtensionTypeInput.tsx
```diff
@@ -1,17 +1,26 @@
...
+  disabled,
+}) => {
   const defaultType: SelectOption = {
     label: "-- select a temporary extension type --",
     value: "",
   };
 
-  return (
+  return disabled ? (
+    <Review heading="Temporary Extension Type">{temporaryExtensionType}</Review>
```
bflynn-cms, line 22: Should we use the passed in temporaryExtensionType to get the label version from: 
export const temporaryExtensionTypes: SelectOption[] = [
  { label: "1915(b) Temporary Extension", value: "1915(b)" },
  { label: "1915(c) Temporary Extension", value: "1915(c)" },
];
That would make it match the format used when viewing the details page. If not its not a big deal and to be honest the label value is a bit redundant but it would be nice if they were kept in sync.

bflynn-cms (approved):
Looks good, made one small comment that can be addressed or ignored as desired! :)

services/ui-src/src/page/temporary-extension/TemporaryExtensionTypeInput.tsx
kristin-at-theta: Honestly, I just thought it looked bad/redundant to have:
**Temporary Extension Type**
1915(b) Temporary Extension

bflynn-cms: I dont disagree, I hate it too. Im guessing that is just what the original story declared for detail view. I part of me just twitches when they arent using the same source regardless of what that source is LOL...but yeah no big deal

PR #1300, opened 2023-07-07
OY2 24488: Data not Cleared with SEA Tool Update

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-24488

### Details
Code successfully creates the correct dates when they first are entered, and also updates properly if the dates change, but does not recognize if a date has been removed. The details page for the package still contains the Final Disposition Date and the Approved Effective Date from the old SEA Tool event.

### Changes
- converted the proposedEffectiveDate field to use emptyField rather than "none" (for consistency)
- added "else" actions to reset the approvedEffectiveDate and finalDispositionDate if they are no longer numbers
- updated the UI to handle both emptyField and dates in the response

### Test Plan
1. Login as a OneMAC User that can view packages
2. Find a package with an Approved Effective Date
3. Use the AWS console to update the latest SEA Tool Event and remove the STATE_PLAN->APPROVED_EFFECTIVE_DATE and/or STATE_PLAN->ACTUAL_EFFECTIVE_DATE
4. Verify the Approved Effective Date is removed from the package
5. Find a package in a Final Status (Package Withdrawn, Approved, Disapproved)
6. Use the AWS console to update the latest SEA Tool Event and change the status to a non-final status (like: Pending)
7. Verify the Final Disposition Date is removed from the package


bflynn-cms (approved):
lgtm!

PR #1309, opened 2023-07-18
Show data filtering options as chips above PortalTable

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23426

### Details

Adds a tray of chips displaying filtering options applied (additive) or removed (subtractive) from the table.

Subtractive filter types are filters that come with all options pre-applied, and allow the user to reduce the number of applied options. Examples: Type, Status

Additive filter types are filters that come with no options pre-applied, and allow the user to reduce visible data by adding a condition. Examples: Territory, time-based filters

Filters are displayed as either `Hidden [Filter]: [Value]` or `[Filter]: [Value]` depending on if they're subtractive or additive.

Screen Shot 2023-07-20 at 3 12 46 PM

### Changes

- Adds Chip, FilterChipTray
- Enumerates column IDs

### Implementation Notes

- One way we can improve the current implementation is to refactor the UI-friendly names for statuses out of `app-api` and into `common`

### Test Plan

1. Login as a State user
2. Navigate to the Dashboard
3. Apply filters by clicking `Filter` and adding/removing options for each
4. Ensure the filter chips show the filters you have applied
5. Ensure each filter can be removed with the `X` icon next to its value
6. Repeat steps 3-5 on the Waivers tab
7. Login as a CMS user
8. Repeat steps 2-6
9. Finally, ensure refreshing and log-out behaviors are still to persist and wipe applied filters respectively


services/ui-src/src/components/PortalTable.tsx
```diff
@@ -50,6 +55,191 @@ const defaultColumn = {
...
+// Statuses shared by CMS and States
+const commonStatuses = ["Approved", "Disapproved", "Package Withdrawn"];
+// Statuses exclusive to CMS
+const commonCMSStatuses = [
+  ...commonStatuses,
+  "Pending",
+  "Pending - RAI",
+  "Pending - Approval",
+  "Pending - Concurrence",
+  "Submitted - Intake Needed",
```
kristin-at-theta, line 83: There was a requirement for state users to not see the CMS statuses - which is why they are converted in app-api.  While bringing the status lookup up into the UI exposes the CMS statuses, it at least does not show when a package is in one of the ones States aren't supposed to see.
I think the filter was data-driven, so only worked with statuses that were in the data.  Not sure if there is a possibility to do that here...

kristin-at-theta (approved):
Tested and everything looks and works quite well, even when I tried to trip it up - nice work!

I'm not a fan of the statuses being hard-coded in the filters... they used to be data driven - which meant:
- state users never saw the CMS statuses
- stories that add new statuses (like the ones Brandon and I created last sprint), get the filter work for free.

But just because I don't like it, doesn't mean it is wrong, so Approved!

services/ui-src/src/components/PortalTable.tsx
kevinhaube: I don't follow -- I've tested as a state user and none of the CMS statuses are visible. This isn't being used for anything except filter adding/removing logic that lets us show which values of `type` and `status` are hidden instead of which are applied ([core function](https://github.com/Enterprise-CMCS/macpro-onemac/pull/1309/files#diff-10c511453a78a20372c19e5174d4c393a67df6b7f85a5c56714b8c540a9376dcR117-R126), [application](https://github.com/Enterprise-CMCS/macpro-onemac/pull/1309/files#diff-10c511453a78a20372c19e5174d4c393a67df6b7f85a5c56714b8c540a9376dcR211-R220)). I do agree it's a redundancy, so I left a TODO about how we could move the mappings of those filter names to the `cmscommonlib` from where we actually map them in the api directory.

kevinhaube: I did find a bug, however, related to this. Not all the 1915 waiver types are represented in the array of types right now, so some chips do not appear when deselecting the waiver type from filters to hide it. 😂 Something for me to fix today!

kristin-at-theta: cmscommonlib is visible as well - I am aware this is one of my tedious spent-too-much-time-fighting-web-crawlers idiosyncrasies, AND I'm also aware we break this all the time.  I feel compelled to mention it when I see it, though.
Screenshot 2023-07-24 at 9 31 14 AM

PR #1310, opened 2023-07-19
Add encryption keys to at rest topics

Dark-Knight-1313:
This push is to add encryption keys to at rest topics

### Details

### Changes

### Implementation Notes

### Test Plan

1. Test step 1
2. Test step 2

### DIFF
--- a/services/app-api/resources/cloudwatch.yml
+++ b/services/app-api/resources/cloudwatch.yml
@@ -4,6 +4,7 @@ Resources:
     Type: AWS::SNS::Topic
     Properties:
       DisplayName: SNS Topic for Monitoring OneMac API lambda errors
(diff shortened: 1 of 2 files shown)


kristin-at-theta (changes requested):
should these be stored as repository secrets or is it ok to have them viewable as plain text in our public repository?  

Also.... the tests that are failing is a little weird - like the code has reverted but the tests are the new tests?  For instance, the TE failure is because in the Filter, the checkbox is only labelled "Temporary Extension" and not "1915b Temporary Extension"

Change Requested: It may be that this branch needs to have origin/develop merged into it, but I feel like we have green test runs now, and the changes made should not have affected that... so those should also be green....

kristin-at-theta (approved):

PR #1314, opened 2023-07-20
Oy2 24562 - cms enable rai withdraw

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-24562

### Details

This story adds an action to CMS and a submit-able form. These are two new things that the existing structure did not support in many ways.

### Changes

- Added new status for WITHDRAW_RAI_ENABLED
- Added new form enableRaiWithdraw and corresponding config
- Added new lambda to support backend api call for enableRaiWithdraw
- -Note that this api does not use submitAny as other State forms do because we decided not to make this an official "event"
- Moved actions out of UI form config and into backend logic in utils/actionDelegate.js; this method is used in other api calls such as getDetail and getMyPackages such that each package returned will contain the list of actions that can be performed based on the caller of the API method and the current package state

### Implementation Notes

- While I moved action logic for packages to the backend, i did not incorporate into package builder as some of the logic also relies now on user roles and must be determined based on who made the api call

### Test Plan

1. Locate a package in RAI Response status
2. Login as state user and submit a formal RAI response if needed
3. Update backend database with fake seatool event to set status of package to Under Review (Pending)
4. Login as cms user (system admin works) and verify the package has the new action for cms
5. Click the action and submit the form
6. Verify Admin section shows change log and in the backend the rai response event record has a status of Under Review (was previously Submitted)


kristin-at-theta (approved):
Approved!

PR #1323, opened 2023-07-25
Dependabot: update packages as needed

kristin-at-theta:
Story: Dependabot

### Details
Packages need to be periodically updated to reflect newer versions.

### Changes
- ran npm audit fix everywhere to catch minor updates
- updated aws-amplify in ui-src...
- updated the repos pointing to CMSgov to now point to new location: Enterprise-CMCS

### Test Plan
1. branch should deploy
2. tests should not fail
3. application should run as before

### DIFF
--- a/.github/oidc/develop.json
+++ b/.github/oidc/develop.json
@@ -1,7 +1,7 @@
 [
   {
     "ParameterKey": "SubjectClaimFilters",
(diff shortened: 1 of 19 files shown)


kevinhaube (approved):
✅ Looks good and functions as expected!

PR #1324, opened 2023-07-25
Update FAQ section

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-24769

### Details

This PR delivers requested updates to the Onboarding Materials FAQ section

### Changes

- Removed **OneMAC State Administrator Guide**
- Changed title from **OneMAC State Submitter Guide** -> **OneMAC State User Guide**

### Implementation Notes

### Test Plan

1. Navigate to `/faq` and ensure the changes are present in the Onboarding Materials section


kristin-at-theta (changes requested):
So, story says: "OneMAC State Administrator Guide should be removed."

Which would not be a problem except that this link works:
https://d19cq2qtxak0yj.cloudfront.net/assets/onboarding/OneMACStateAdministratorGuide.pdf

So, if they have bookmarked it, they can still see it, which I feel is contrary to the AC

kristin-at-theta (approved):

PR #1331, opened 2023-08-01
OY2 23845: Withdraw Formal RAI Response - Email

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23845
Endpoint: https://d1jfwoxkadkgby.cloudfront.net

### Details
This feature creates the CMS notice and the State receipt for the RAI Response withdraw action.

### Changes
- updated package builder to include the cpocEmail and the reviewTeamEmails in the package item
- added the State and CMS Withdraw RAI response emails
- updated the CMS email creation to include the async lookup of the cpoc and srt emails
- because the current withdraw rai reponse form is type-agnostic, added a type-check to determine which CMS emails are included.
- Updated seed data to better support automated testing
- Updated resetData to better support automated testing.

### Test Plan
1. Find a way to see/be included in the emails being sent.  (using different ways to verify might uncover missed bugs/acs)
2. Login as a State user
3. Navigate to dashboard
4. perform 5-9 for each of: Medicaid SPA, CHIP SPA, Initial B waiver, B waiver Renewal, B waiver Amendment, C Waiver "App K" amendment.
5. Locate a Medicaid SPA in the "RAI Response Withdraw Enabled"
6. Begin the "Withdraw Formal RAI Response" action
7. Submit the form
8. Verify the state receipt is sent to the correct parties and looks as described in the story AC 
9. Verify the CMS notice is sent to the correct parties and looks as described in the story AC


services/app-api/resources/ssm-params.yml
```diff
@@ -21,44 +21,44 @@ Resources:
...
 
   CmsDhcbsooEmail:
     Type: AWS::SSM::Parameter
     Properties:
       Type: String
       Description: Branch specific email addresses of the DHCBSOO group for CMS
       Name: !Sub /configuration/${self:custom.stage}/email/cms_dhcbsoo_email
-      Value: "OneMAC@cms.hhs.gov"
\ No newline at end of file
+      Value: "\"DHCBSOO group\" <OneMAC@cms.hhs.gov>"
```
bflynn-cms, line 64: I think this exposes a flaw we have known about with the configuration of these emails in SSM and using defaults in lower environments such that by changing the value here it will rebuild the SSM parameter and we will have to manually set the proper values again when it gets to PROD.

I do realize the change was likely so you could tell which default email was actually being assigned since there are now multiple groups being CC'd, so I do like the change.

services/app-api/email/stateWithdrawalReceipt.js
```diff
@@ -51,10 +51,9 @@ export const stateWithdrawalReceipt = async (data, config, user) => {
     ToAddresses: [],
     CcAddresses: [],
     Subject: `${config.typeLabel} Package ${data.componentId} Withdraw Request`,
-    HTML: `
-        <p>The OneMAC submission portal received a request to withdraw a package. You are receiving this email notification as ${
-          data.componentId
-        } was withdrawn by ${user.fullName} (${
+    HTML: `<p>The OneMAC submission portal received a request to withdraw a package. You are receiving this email notification as ${
+      data.componentId
+    } was withdrawn by ${user.fullName} (${
```
bflynn-cms, line 56: Just double checking but I think I only see whitespace changes here?

bflynn-cms (commented):
Just leaving code comments for now but have not finished testing.

Looks good! Made one pertinent note regarding resetting of emails in SSM.

services/app-api/resources/ssm-params.yml
kristin-at-theta: Oh.... great point and that totally sucks :( - and yes, I did it so we could tell which actual email variable was in play because they are all the same default ...

bflynn-cms (approved):
Everything looks good to me.  I like the way you built in to distinguish the many default emails that are the same in non prod environments!

Approved!

PR #1333, opened 2023-08-02
Wrap long status text in column

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-24499

### Details

This PR updates the `Status` column styling to accomodate longer statuses like `Formal RAI Response - Withdrawal Requested`

### Changes

- Applies new `status-bandage` class to status column **only**.
- Has a new max and min width so there is still grow/shrink room when scaling to smaller screens

### Implementation Notes

### Test Plan

1. Ensure longer statuses wrap in the Status column


kevinhaube:
https://d2t8ivqk214ipf.cloudfront.net/

services/ui-src/src/components/PortalTable.tsx
```diff
@@ -373,6 +373,12 @@ export default function PortalTable<V extends {} = {}>({
                     {row.cells.map((cell, index) => {
                       return (
                         <td
+                          className={
+                            headerGroups[0].headers[index].id ===
+                            "packageStatus"
+                              ? "status-bandage"
+                              : ""
+                          }
```
bflynn-cms, line 381: I feel like a better way to do this would be thru the cell property of the columns definition passed to PortalTable rather than enforcing style using a particular column name from within the PortalTable component itself.

services/ui-src/src/index.scss
```diff
@@ -67,6 +67,16 @@ $color-primary-alt-lightest: #f0fafe;
...
+  .status-bandage {
+    // the min and max allow for minor growing/shrinking
+    // to condense the width of the column when scaled down
+    // At full screen, the extra 25px lets the status breathe
+    max-width: 170px;
+    min-width: 145px;
+    // Responsible for wrapping any column
+    white-space: normal;
+  }
+
```
bflynn-cms, line 79: This works though on large screens still wraps when unnecessary.  Having said that I can also see the wrapping at a fixed size making sense too so that a single column doesnt push the rest too far to the right. Just a note, nothing to really fix here if this was intentional.

bflynn-cms (changes requested):
Left a comment on method for applying class to cell elements. I dont think we should hard code the column name and style into the PortalTable component.

services/ui-src/src/index.scss
kevinhaube: This behavior addresses AC number 1: 
> The status column should be sized reasonably in comparison to the table size and other columns

No official size was given, so I used my better judgement! I did update the method re: the above comment, so now technically the _column_ doesn't have a max width, just the `p` element inside, and the cell will adjust width accordingly.

bflynn-cms (approved):
Looks, good! Thanks for making those changes!

Approved.

PR #1340, opened 2023-08-11
Consolidate uses of card styling to single component

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-23648

### Details

This PR consolidates all use of a box w/ gradient top and gray border as the `MACCard` component system, and replaces prior distributed styles from:
- Home
- Details (two at the top, and used to wrap file lists)
- New Submission triage option
- Request Role
- User Page state access cards
- FAQ contact card

### Changes

- Consolidate styling and layout code in `MACCard.tsx`
- Unit tests for new components
- Replaced uses around application

### Implementation Notes

- ~~🚨 **This changed a split pattern** - we were using both `onClick` and `linkTo` to navigate within triage options (now called MACFieldsetOptions) and now we ONLY rely on `linkTo`. Be sure to double check the Request Role behaviors and New Submission triage behavior. To the best of my abilities, I've manually tested on my local, as well.~~
> This created a bug. The onClick is needed only in one case, and comments were left to help with later refactoring.
### Test Plan

1. === HOME ===
2. Ensure the Home Page MACCards display and scale properly
3. === PROFILE ===
4. Navigate to User Proflie
5. Ensure the State Access cards render properly
6. Add a state to your user's access and ensure the card appears
7. Click the X on the new pending state to revoke the request and ensure this behavior works
8. Navigate to Request a Role
9. Ensure the option given reflects the proper role to be requested by that user-type
10. Ensure the option links to the proper page and the page renders with no errors
11. === NEW SUBMISSION ===
12. Navigate to the Dashboard
13. Click New Submission
14. Ensure the triage options display and link to the proper locations
15. Navigate back to the dashboard
16. === PACKAGE DETAILS ===
17. Click on a package until you find one with attachments
18. Ensure the cards at the top of the details page display properly
19. Ensure the file list is displayed in a card properly


kristin-at-theta (approved):
Thank you for cleaning this up and making it a proper design component!

I did a visual and functional comparison between "then" and "now" and could not find any functionality differences and did see some spacing/sizing differences, but found I preferred the "new" in each case, so good job!

a little confused about the "withGradientBar" prop - are there any that do not have the gradient?  I thought maybe the "State Access" card on the user profile page, but when I inspect that, it is not a MACCard.  Anyway, just something I wondered.

Approved!

kevinhaube:
@kristin-at-theta Thanks!! :D I found the new one has more consistent padding/margins so they look more uniform than prior which might be why we both prefer them. As far as the `withGradientBar` prop, the only place using it right _now_ is the `MACFieldsetCardOption` component which actually looks like its left over from my first round of attempts. There's no reason it can't just be replaced with a `div` which would render the `withGradientBar` prop useless! :) I'll refactor that little bit out today; good catch.

PR #1344, opened 2023-08-21
Fixes filters by date to include full days

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-25078

### Details

Applies `startOfDay` and `endOfDay` function to the date filter so all results from both days are included.

Dates from the “Past 7 Days” or “Past Month” selections are normalized to “startOfDay” and “endOfDay” values and work as expected, but dates hand-selected by the range picker weren’t. They had the exact time of the selection shown, so effectively filtering for Aug 2-17 when at 10:43am would filter Aug 2 10:43am - Aug 17 10:43am.

```
Current
0: "2023-08-02T14:43:09.640Z"
1: "2023-08-17T14:43:09.640Z"
```
```
Fix
0: "2023-08-02T00:00:00.640Z"
1: "2023-08-17T23:59:59.640Z"
```

### Changes

- Filter value is now checked for length (comes in as an array of 2 date strings in [earliest, latest] order) and if the values are present, the function will normalize the times and return the normalized values

### Implementation Notes

- This is how we handle the same behavior with our "Past 7 Days" type of option

### Test Plan

1. Set any date filter on the Dashboard
2. In dev tools, go to Application
3. Ensure the filter local storage shows the values with normalized time reflecting the fixed values above


kristin-at-theta (changes requested):
Wow, nice bug hunting!  Crazy that it takes the time of the selection as the time, but I guess if you don't specify, have to use *something* 

So... I can't test the application itself because the deployment failed in the configure, so there are no test accounts to use.... I also see the latest deployment push had an issue with the linter (looks like one where a restart will fix that)

Please get a completed deployment made so I can test :) thanks!

kevinhaube:
https://d3bdaogerbfxff.cloudfront.net/
@kristin-at-theta Looks like AWS chose to work today; should be good for review now

kristin-at-theta (approved):

PR #1345, opened 2023-08-22
OY2 23552: Latest Package Activity - Package Details Page + Dashboard

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-23552

### Details
Knowing the last OneMAC Action taken is something the users want to see.   It is called the "Last Package Activity"

### Changes
- added the lastActivityTimestamp attribute to the Package Item
- included it in the getMyPackages projection
- set the value in the package builder according to the story AC
- added the column to the dashboard
- added the field to the details pages in the assigned spots
- added the submission ID field to the details pages, where requested.

### Test Plan
1. Login as a 
2. Test step 2

### DIFF
--- a/services/app-api/getMyPackages.js
+++ b/services/app-api/getMyPackages.js
@@ -39,7 +39,7 @@ export const getMyPackages = async (email, group) => {
         ExclusiveStartKey: null,
         ScanIndexForward: false,
         ProjectionExpression:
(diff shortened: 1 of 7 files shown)


bflynn-cms (approved):
Code looks good.

Testing looks good and dates update as expected as new actions are taken. 

I did notice that we are no longer hitting our 1024 goal of showing all columns due to the new column + long column titles. We might want to consider wrapping column titles at some breakpoint. Now at 1024 the actions column is hidden. This may just be an inevitability though as CMS decides to show more and more columns on the dashboard.

Approved.

PR #1354, opened 2023-08-30
OY24856 - rai response attachments

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-24856

### Details

Allow same attachment types from original submission as optional attachments for subsequent submissions on RAI Responses.

### Changes

- Added waiverAuthority as property on waiver rai response forms so that it can be passed into submit
- Added passing of waiverAuthority and waiverParentType in location state in rai response action
- Added method in waiver rai response config in common to determine conifg based on authority and type
- Updated WaiverRaiForm to compile formConfig based on waiver type and waiver authority using common method
- Updated Medicaid, and Appk configs to override optional attachments using parent config attachments
- Updated submitRaiResponse to use config based on waiver type and waiver authority using common method

### Implementation Notes

- I used a slightly different pattern to determine proper form config for waivers such that we didnt need an individual form and submit method per waiver type and waiver authority

### Test Plan

1. Login as state submitter
2. Find a package in RAI Response status (Medicaid Spa, Initial Waiver (B/B4), Renewal Waiver (B/B4), Amendment Waiver (B/B4), Appendix K)
3. Select Respond to RAI action
4. Verify attachments have the one required attachment = RAI Respone and optional attachments = original submission attachment types
5. Submit and verify detail page for the package has the RAI Response with the attachments


kristin-at-theta (changes requested):
I definitely like how you used the existing required and optional attachment lists to build the new ones for the RAI.

I'm not a fan of how the waiver RAIs now break the pattern of "one form/one config," but in some ways, it doesn't matter anymore, we're not going to really have to maintain this.  Yes, I know the waiver authority separation did it, too.  Wasn't a fan then, either, but - eh.  It bothers me enough to whine about it, but its really a style thing and this is fine/works and all that. 

I changed some of the waiver authority values in the data to see the different forms and notice that there are some waiver packages that are tagged as 1915(b) but have the 1915(b)(4) attachment types... sigh.

Change Requested: Is the CMS 179 upload on the Medicaid SPA form still supposed to be restricted to a single upload? I see that it is restricted on the form, but the attachments text doesn't say anything about it (and neither does the story AC) - Please make it consistent between the text and the restriction.

bflynn-cms:
> 

@kristin-at-theta 
I fixed the attachment language so that it referenced the CMS 179 behavior and is shared with the main medicaid spa form.

As for the form/config, I did intentionally deviate because I wanted to explore other ways. I felt our current way was getting too bloated. Maybe this way is good/bad or just bad because there are now two ways but I wanted to try. If you feel strongly enough about it we can split by type (medicaid, spa, initial waiver, etc) but there is no actual form for b vs b4 waivers. We could make that happen too and rai component link would be dynamic based on waiver authority but then I think we should just have an overall refactor story for all of it.

I am not able to replicate what you are seeing with waiver rai responses having wrong attachments based on authority type. Maybe we can sync up on this and have a wakthru?

kristin-at-theta (approved):

kristin-at-theta:
> I am not able to replicate what you are seeing with waiver rai responses having wrong attachments based on authority type. Maybe we can sync up on this and have a walk thru?

Explained in a huddle - its that the test data has the conflicts - the application is fine and we expect the PROD data to also be fine.

PR #1364, opened 2023-09-08
Filter Chip Refactor: No longer reliant on hard-coded statuses and array diffing

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-25096

### Details

This PR overhauls the filter chip logic in a few notable ways.

1. Handling of "additive" and "subtractive" filters no longer requires diffing a hard coded array of possible `componentType` or `currentStatus` values (this was the origin of the bug in 25096)
2. Due to complications with filter management and react-table, filter chips decouple their logic enough to manage themselves via `FilterChipProvider`, a React context with intrinsic state and processing logic. The chips still interact with filter state, but are not reliant on it as a source of truth for which chips to render. This means we have `setFilter` and `updateFilterChip` logic whenever we set filters via the UI.
3. Like number 2, we **also** store saved state separately as processing filter saved state to re-render chips on refresh or when swapping tabs proved impossible due to the way Checkbox filters store the _checked_ values rather than the hidden values. Relying on filter saved state alone would've required the same array diffing that caused this bug in the first place.

### Changes

- Extract `FilterChipTray`
- Create `FilterChipProvider` context and wrap the filter UI and chip tray UI so each can access and update the chips in the tray
- Save and load filter chip state from localStorage

### Implementation Notes

- This approach decouples the filter chip display from the intrinsic filter state of the table. However, through the filter UI and the chip UI, we are wired up to add and remove objects in sync so there should be no discrepancies between filters set and shown in the filter UI, and the chips displayed in the chip UI.

### Test Plan

1. Log in as a CMS or State user -- both should have access to the statuses mentioned in the source ticket.
2. Ensure that each status and type are able to be rendered in a chip
3. Set some filters, and swap tabs and/or refresh the page
4. Assert the chip and filter configurations both are loaded properly upon page load
5. Log out and back in
6. Assert the chips and filter configurations are reset upon logout (this can also be done by viewing the saved state in localStorage


kevinhaube:
Lol, oops. Thought I'd marked ready for review yesterday 😂 my bad!

kristin-at-theta (approved):
I quite like the use of components and contexts to logically group the different parts of the search and filter section.

I see the filters are sent to the tray using a prop... would structuring them as children of the tray be better?  (that's just me wondering, I am amazed by your React skills :) )

There's a lot of logic cleanup and restructuring here and it looks like a much more stable design.  I am not strong enough in React to be able to "see" if there are any issues/mistakes, but I do find that I can usually test things in a way that shows if there's unstable logic at work.  Testing showed this implementation to be very clean and well-behaved.

Nice work!  Approved!

kevinhaube:
@kristin-at-theta Great question! We don't actually render based on props (except that silly record count value, but alas). We're iteratively rendering the children of the tray using the context's state that we hook into from `FilterChipTray`. The `filters` array you see passed as a prop is exclusively used on [lines 46-59](https://github.com/Enterprise-CMCS/macpro-onemac/pull/1364/files#diff-ff49a360f41a74f1100e9a92714f554d6417c7845582f8178ad9afbb453c2434R46-R59) as a way for us to alter the filter value, removing a specific filter, when we hit `x` on the chip!

PR #1366, opened 2023-09-14
Update additional info for Enable RAI Withdraw form

kevinhaube:
Story: https://qmacbis.atlassian.net/browse/OY2-25312

### Details

Updates the additional info text for Formal RAI Response Withdraw form

### Changes

- `TextArea.hint` allows for a `ReactNode` to be passed in so I expanded the `OneMACFormConfig.addlInfoText` property to use the same type

### Implementation Notes

- N/A

### Test Plan

1. Locate a package with Enable Withdraw Formal RAI Response action available
2. Click the action and view the subsequent form page
3. Ensure the text matches spelling and formatting of the AC

### DIFF
--- a/services/ui-src/src/libs/formLib.tsx
+++ b/services/ui-src/src/libs/formLib.tsx
@@ -24,7 +24,7 @@ export type OneMACFormConfig = {
   attachmentsTitle?: string;
   attachmentIntroJSX: JSX.Element;
   addlInfoTitle?: string;
-  addlInfoText?: string;
+  addlInfoText?: string | React.ReactNode;
   addlInfoRequired?: boolean;
   requireUploadOrAdditionalInformation?: boolean;
   landingPage: string;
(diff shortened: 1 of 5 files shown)


kristin-at-theta (approved):
All (most?) of the other formatted static texts use JSX Elements rather than React.ReactNode.  I can't say I know much of how that is different, but I certainly prefer having only one way to maintain.

I also find it a little weird that the color is not the same color as the bolded text on other forms - it's somehow grayer?

That being said, this does match the story AC, so, Approved!

kevinhaube:
@kristin-at-theta 

#### React's typing
The [official docs](https://react.dev/learn/typescript#typing-children) outlining `ReactNode` and `ReactElement` do a better job than I typically do in describing it, but the core of it is that a JSX element is a `ReactElement`, and a `ReactNode` expands the possibilities from only that JSX to JSX or any JavaScript primitive type, which is why we can just pass in strings in to the component like we have historically. Using `ReactNode` was more of a way to match the component's possibilities, but we could limit it to `string | JSX.Element` and still have completely valid properties for our config.

#### Boldness styling
I think this might be because we're setting things like `font-weight` in CSS to achieve boldening via classes instead of via markup, and in this case we're using the designated markup tag instead of plugging in a CSS class that has text style modifications

PR #1368, opened 2023-09-21
Oy2 23799 - add emails for subsequent submissions

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-23799
Endpoint: https://dq6gyijpx3ouv.cloudfront.net/

### Details

Add state and cms emails to subsequent submission workflow

### Changes

- Extracted SRT and CPOC email collection from withdraw rai email to common method so that it could also be used in subsequent submission emails
- added state and cms email templates + tests
- added email templates to default subsequent submission config

### Implementation Notes

- I followed ACs and put the only recipients of the CMS email to be CPOC and SRTs but worry that there could be a package without a CPOC or SRT from SeaTool? In which case the email would fail.

### Test Plan

1. Log in as state submitter
2. Locate package of any type in Under Review status
3. Submit subsequent submission documents
4. Verify emails

### DIFF
--- a/services/app-api/email/CMSSubsequentSubmissionNotice.js
+++ b/services/app-api/email/CMSSubsequentSubmissionNotice.js
@@ -1,0 +1,50 @@
+import { formatPackageDetails } from "./formatPackageDetails.js";
+import { getCPOCandSRTEmailAddresses } from "../utils/getCpocAndSrtEmail.js";
+
+/**
+ * Package Submission email to CMS
+ * @param {Object} data from the package.
+ * @returns {Object} email parameters in generic format.
+ */
+export const CMSSubsequentSubmissionNotice = async (data, config) => {
+  const CMSEmailItem = await getCPOCandSRTEmailAddresses(data.componentId);
+
+  const ToAddresses = CMSEmailItem.reviewTeamEmailList
+    ? [...CMSEmailItem.reviewTeamEmailList]
+    : [];
+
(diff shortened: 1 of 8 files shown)


kristin-at-theta (changes requested):
Oops, somehow lost my "intro" - editing to add...
Nice work with pulling the CPOC and SRT email finder into a util function!  Good observation about if there is no CPOC, too. might want to add that to the story comments?

Change requested (to AC or to code... they don't match): Weirdly, the story says: \<insert attachment type(s)\> in that intro line and what is there is the \<package type\>... I feel like either the story AC should change or should do the attachment types (though we have never included those before for the state user...)

there are a bunch of label differences between the state user email and what is in the story - I feel this is more because the story writer did not know what labels are used in the emails and less that they want them changed...  In any case, since the data is the same, just has different labels, I don't think that's worth customizing the emails to be different here...

the CMS email has extra fields compared to the story, the Submitter Name and Email address - looks like they were purposefully removed this time?  It does seem weird that the email to the state user contains their contact information and the CMS user's email does not...

bflynn-cms:
> https://qmacbis.atlassian.net/browse/OY2-23799

@kristin-at-theta I worked with Jaime and Rachel to get the ACs updated to match current email formats for fields displayed and to update the state receipt email text to be similar to cms email and list the submission type rather than attachment types.

PR #1375, opened 2023-10-02
PI01.02 25326: Add Success Alert for Enable and Disable Withdraw Formal RAI Response Actions

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-25326
Endpoint: [Click here.](https://dk03qsixo2krz.cloudfront.net)

### Details
It was noticed during the demo of 25156 that the alert messages for the enable/disable actions were the default success alerts and that a more informative message would be better.  This is done as a separate branch so as to not interfere with 25156's journey.

### Changes
- Added the new alert messages for enable and disable
- Updated the enable/disable lambdas to return the correct responses for the enable/disable action

### Implementation Notes
- Note that the merge is into 25156 branch so that the alerts are in the branch that will be demoed on Wednesday to the Business.

### Test Plan
1. Login as a CMS user
2. Find a package in "RAI Response Withdraw Enabled"
3. Disable the enable, verify the alert matches the AC
4. enable the package again, verify the alert matches AC.


bflynn-cms (approved):
Looks good!

Approved.

PR #1398, opened 2023-10-26
PI1.2: part 1 of Refactoring Cypress Test Data for Dynamism

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-25593

### Details
While unsuccessful at the larger initiatives around separating out the data... creating the seed-data "tree" will (I hope) begin to solve some of the issues around the very large seed data file and make it so updates to a file won't so easily get overwritten.

### Changes
- updated loadTestUsers.py to use a json file
- broke out one-seed.json into a seed-data directory tree so we can separate the data into the individual test needs
- built a data-glomer to combine all the small files into the large seed-data for github action use (I could not figure out how to get the sources variable to use wildcards and thought listing each file was just asking for trouble)

### Implementation Notes
- the tree is started and the first level is broken out.  Eventually, would like to see each package in its own json file with a descriptive title.  This is in anticipation of using the smaller data files in individual tests.

### Test Plan
1. Branch should deploy
2. seed data should work as before.
3. Local development should also have seed data (might have to build the all-data.json locally)


kristin-at-theta:
That final test failure is a mismatch between a specific data lookup, and the potential use of that same item in a generic script that alters the package being tested to a fail state.  There are a couple solutions for this, but I felt like that was outside the scope of the data breakout. :)

bflynn-cms (approved):
I like how you made a script to combine all json into one for the seed data. Pity it cant natively do a dir path. This way is nice though such that any added files will automatically get picked up instead of having to name each one. Its awesome we now have different categories with data split up by type. I wont miss the 4M line seed file!! 🤣 
Approved!

PR #1401, opened 2023-11-01
PI1.4 26064: CMS RAI Response Withdrawal enablement update

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-26064
Endpoint: [Click here](https://d2pw7kqcao5lf0.cloudfront.net/)

### Details
As a CMS user, I do not need to provide a change reason when enabling or disabling a RAI Response Withdrawal.

### Changes
- removed the Change Reason from the Enable RAI Response Withdraw and the Disable RAI Response Withdraw forms.
- added a new config option to allow forms to not have the text field 
- removed the Change Reason from any Admin Changes that are for Enable/Disable RAI Response Withdraws

### Test Plan
1. Login as a CMS user
2. Navigate to Dashboard
3. Find a package in Pending Status with an Formal RAI Response date (or, a package in Formal RAI Response Enabled)
4. Go into package details and click the "Enable" action
5. Verify the form no longer requires a reason
6. Verify submitting the form Enables (or Disables) the RAI Response Withdraw
7. Verify that the Admin Changes section does not have a Change Reason for these enabling/disabling actions.

### DIFF
--- a/services/ui-src/src/libs/formLib.tsx
+++ b/services/ui-src/src/libs/formLib.tsx
@@ -26,6 +26,7 @@ export type OneMACFormConfig = {
   addlInfoTitle?: string;
   addlInfoText?: string | React.ReactNode;
   addlInfoRequired?: boolean;
+  noText?: boolean;
   requireUploadOrAdditionalInformation?: boolean;
   landingPage: string;
   confirmSubmit?: ConfirmSubmitType;
--- a/services/ui-src/src/page/section/DetailSection.tsx
+++ b/services/ui-src/src/page/section/DetailSection.tsx
@@ -301,12 +301,16 @@ export const DetailSection = ({
                     <Review className="preserve-spacing" heading="Change Made">
                       {adminChange.changeMade}
                     </Review>
-                    <Review
-                      className="preserve-spacing"
-                      heading="Change Reason"
-                    >
-                      {adminChange.changeReason}
-                    </Review>
+                    {!adminChange.changeMade.includes(
+                      "abled State package action to withdraw Formal RAI Response"
+                    ) && (
+                      <Review
+                        className="preserve-spacing"
+                        heading="Change Reason"
+                      >
+                        {adminChange.changeReason}
+                      </Review>
+                    )}
                   </AccordionItem>
                 );
               })}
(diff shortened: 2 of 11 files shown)


services/ui-src/src/libs/formLib.tsx
```diff
@@ -26,6 +26,7 @@ export type OneMACFormConfig = {
   addlInfoTitle?: string;
   addlInfoText?: string | React.ReactNode;
   addlInfoRequired?: boolean;
+  noText?: boolean;
```
bflynn-cms, line 29: not sure i like this name givven that all other fields related to additional information section have the addlInfo prefix.....but really who cares since this is all going away *shrug*

services/ui-src/src/page/section/DetailSection.tsx
```diff
@@ -301,12 +301,16 @@ export const DetailSection = ({
                     <Review className="preserve-spacing" heading="Change Made">
                       {adminChange.changeMade}
                     </Review>
-                    <Review
-                      className="preserve-spacing"
-                      heading="Change Reason"
-                    >
-                      {adminChange.changeReason}
-                    </Review>
+                    {!adminChange.changeMade.includes(
```
bflynn-cms, line 304: why not use the absence of adminChange.changeReason to determine if this section should be shown? Im guessing there is a good reason?

bflynn-cms (approved):
Testing checks out, though you did fool me with the wrong link in the "click here"! Which in hindsight provided for a good negative test before testing the real branch! LOL

Made a few comments but nothing really "needing" to be changed.

Approved!

services/ui-src/src/libs/formLib.tsx
kristin-at-theta: Yeah, I struggled with the name, for some reason couldn't bring myself to say addlInfo when the label/fieldname for the one to remove is "Change Reason."  I ended up saying noText because it was short and was going to use a better name later and well... it's short at least?

services/ui-src/src/page/section/DetailSection.tsx
kristin-at-theta: I thought about that and decided that No Change Reason on an entry that could have one is information someone might want.  I know that's kind of silly, but not like it was that hard to properly identify the enable/disable ones (even without a special type field :) )

PR #1402, opened 2023-11-03
Oy2 26063 - RAI Response Withdraw Enabled is no longer a package status

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-26063

### Details

RAI Response Withdraw Enabled is no longer a package status. When the enable withdraw action is taken the package status should remain in Pending/Under Review. Instead there will be a "substatus" displayed under the status on both the dashboard and the details page.

### Changes

- removed newStatus from enable and disable withdraw config since they are not actually used
- removed setting the eventTimestamp and refactored buildAnyPackage to not rely on that
- refactored buildAnyPackage to detect substatus and lock in package status

1. what fields are new
The package item will now have a new field named substatus. This will drive action logic and display of substatus on pacakge details and package list screens
2. what fields are removed
None
3. what fields are going to hold different content
package status will no longer be able to hold "RAI Response Withdraw Enabled" status and will instead stay in Pending

### Test Plan

1. Login as CMS user
2. Find any package in Pending status with a submitted RAI
5. Click the Enable Rai Withdraw action
6. Verify package status remains Pending and substatus shows underneath
7. Navigate back to package list and verify package has both status and substatus in table
8. Navigate back to detail view for package and select disable withdraw rai response
9. Verify package still remains in original status (Pending)
10. enable withdraw rai again
11. login as state user and request withdraw
12. verify status changes to xxxxxxxxx


kristin-at-theta (changes requested):
Change: Story AC shows the 2nd clock message AND the enabled substatus in details page card for CMS users

Change? substatus on dashboard and details page has different styles than in branch.

I don't care for the ENABLED and DISABLED being left in ONEMAC_STATUS, since they are no longer OneMAC Statuses.

Not sure why currentStatus is being forced to the previous package item's status, I don't see that requirement in the AC.  I did test that it appears to be working as intended, I added a SEA Tool event to change the package status and it stayed in Pending until I disabled the enable and then switched to new status.

I think I would've dropped subStatus all the way down to the events, so we're not mucking with the anEvent.currentStatus and all that timestamp nonsense jury-rigged in there when statuses became mutable.  BUT, the way you did it involves less mucking with critical path code, so eh.

I feel like only the missing 2nd clock banner should be addressed.  The other comments might be more relevant if this application had to live longer than the next few months, but.... makes sense to choose short-term reliability over long-term readability/future stability when long-term is unlikely

bflynn-cms:
> Change: Story AC shows the 2nd clock message AND the enabled substatus in details page card for CMS users
This was really part of OY2-24635 since previously the latestRaiTimestamp was removed as soon as the withdraw was enabled but now needs to stay until the state requests withdraw. But I do understand its part of the total workflow and was in the screenshots. I ultimately decided to just add the changes needed for oy2-24635 to this branch.
> 
> Change? substatus on dashboard and details page has different styles than in branch.
There is no formal Figma or design for font/text of the substatus. During refinement Jaime even noted he just threw it in the attached images but we didnt need to follow that specific format. We can see what they think on demo and adjust from there.
> 
> I don't care for the ENABLED and DISABLED being left in ONEMAC_STATUS, since they are no longer OneMAC Statuses.
We actually are still using the WITHDRAW_RAI_ENABLED status on the RAI (just no longer on the package level). Im not sure there ever was a DISABLED status?? So not sure if you are referring to something else.
> 
> Not sure why currentStatus is being forced to the previous package item's status, I don't see that requirement in the AC. I did test that it appears to be working as intended, I added a SEA Tool event to change the package status and it stayed in Pending until I disabled the enable and then switched to new status.
This was done to keep the previous intent to not allow a change to overall status from SEATOOL events once the RAI withdraw has been enabled until its either disabled or withdrawn. Since we no longer have the package level enabled status to lock at I figured we should lock on the existing package status at time of enable which at this point in time will always be PENDING.
> 
> I think I would've dropped subStatus all the way down to the events, so we're not mucking with the anEvent.currentStatus and all that timestamp nonsense jury-rigged in there when statuses became mutable. BUT, the way you did it involves less mucking with critical path code, so eh.
This is true....I could have made the enable/disable/withdraw rai events set the substatus and use that to drive package level substatus....in hindsight that probably would make more sense.
> 
> I feel like only the missing 2nd clock banner should be addressed. The other comments might be more relevant if this application had to live longer than the next few months, but.... makes sense to choose short-term reliability over long-term readability/future stability when long-term is unlikely

kristin-at-theta (approved):

kristin-at-theta:
Changes look good, thank you for the explanations :) Tests look good!  Approved!

PR #1403, opened 2023-11-03
PI1.4 26079: FAQ Search expansion

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-26079

### Details
Add an "Expand all" button to the FAQ page to allow Ctrl-F to search all FAQ text.

### Changes
- moved the FAQ content into the React state to get native rendering
- added isOpen to the faqcontent items in json
- switched the Accordian component from default to Controlled open to allow the application to open and close the questions
- added the handler for opening and closing
- added a button at the top of the page that opens all the questions

### Implementation Notes
To do this properly, would really be best to remove the hashHandler and use the state to manage which item shows, etc.  Also, there is probably value in dividing the components even further, so a change to one QandA doesn't have to alter the entire faq content state to render.  HOWEVER, this was only a 2 point story and it does not look like our work on the FAQ page is going to be migrated.

Also note... using the Expand button does NOT automatically re-trigger the Ctrl-F search in the browser - you have to edit the text to be searched in some way.

### Test Plan
1. Open the application
2. Click the FAQ link
3. Click the "Expand all to search with CTRL+F" button
4. Verify the all sections expand
5. Verify text is searchable
6. Verify other application FAQ links work as before (expanding and moving to correct places)

### DIFF
--- a/services/common/routes.js
+++ b/services/common/routes.js
@@ -10,6 +10,7 @@ export const ROUTES = {
   DETAIL: "/detail",
   USER_MANAGEMENT: "/usermanagement",
   FAQ: "/FAQ",
+  FAQ_OPEN: "/FAQ#open",
   FAQ_ACCEPTED_FILE_TYPES: "/FAQ#acceptable-file-formats",
   FAQ_ATTACHMENTS_MED_SPA: "/FAQ#medicaid-spa-attachments",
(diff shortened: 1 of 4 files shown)


bflynn-cms (approved):
Everything looks and functions well!

Approved

PR #1412, opened 2023-11-20
Develop Package fixes

kristin-at-theta:
### Details
The hot fixes caused develop and master (val) to get out of sync.  When merging back, the packages got put into a state where unit tests and automated testing stopped working.  This branch *should* fix this problem!

### Changes
- some packages got updated that should not have
- reconfigured the package-locks to match the correct package files

### Test Plan
1. Branch should deploy
2. Application should run
3. Tests should succeed/fail as before


bflynn-cms (approved):
Not sure how you figured out what to fix!
Its running now though!
Approved

PR #1413, opened 2023-11-22
PI1.05 26321: Add Final Disposition to the Dashboard

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-26321

### Details
CMS has asked that Package Withdrawn date get added to the dashboard for use with filters and reports.  We discovered that "Final Disposition" date is the Package Withdrawn date when the package is in Withdrawn status, so decided to add that column to the dashboard instead.

### Changes
- Because of weird date conversion issues, see "Implementation Notes," added the finalDispositionTimestamp to the package items and added enough milliseconds to be on the correct date in the eastern timezone.
- Added finalDispositionTimestamp to the dashboard as a hidden column.
- added filter details to make filters work as dictated

### Implementation Notes
The date being used is the SEA Tool STATUS_DATE on the STATE_PLAN record.  Now... that date is captured in the SEA Tool application as a mm-dd-yyyy date and sent to us as a timestamp... however, the timestamp is for GMT, so in any date conversion that uses local timezone, it is a day off.  We get around this in other places by converting the date on the back end (which uses GMT), but in order to use the date filter widget, we need a timestamp, so I've added enough milliseconds to the "timestamp" to have it be on the correct date in either GMT or Eastern.

### Test Plan
1. Login as a state user
2. Navigate to Dashboard
3. Verify Final Disposition is not a default column
4. Click "Show/Hide Columns"
5. Select Final Disposition
6. Verify the column is added to the dashboard and contains the correct value (can compare to details page)
7. Verify filters work as expected
8. Verify saving of filters works as expected
9. Verify exports are correct
10. Do the same tests with a CMS user

### DIFF
--- a/services/one-stream/package/buildAnyPackage.js
+++ b/services/one-stream/package/buildAnyPackage.js
@@ -256,17 +256,12 @@ export const buildAnyPackage = async (packageId, config) => {
         if (seaToolStatus && SEATOOL_TO_ONEMAC_STATUS[seaToolStatus]) {
           const oneMacStatus = SEATOOL_TO_ONEMAC_STATUS[seaToolStatus];
           putParams.Item.currentStatus = oneMacStatus;
-          if (finalDispositionStatuses.includes(oneMacStatus)) {
-            putParams.Item.finalDispositionDate = DateTime.fromMillis(
-              anEvent.STATE_PLAN.STATUS_DATE
-            ).toFormat("yyyy-LL-dd");
-            // have to add 18000000 seconds because of time zone
-            putParams.Item.finalDispositionTimestamp =
-              anEvent.STATE_PLAN.STATUS_DATE + 18000009;
-          } else {
-            putParams.Item.finalDispositionDate = emptyField;
-            putParams.Item.finalDispositionTimestamp = 0;
-          }
+          putParams.Item.finalDispositionDate =
+            finalDispositionStatuses.includes(oneMacStatus)
+              ? DateTime.fromMillis(anEvent.STATE_PLAN.STATUS_DATE).toFormat(
+                  "yyyy-LL-dd"
+                )
+              : emptyField;
         }
       }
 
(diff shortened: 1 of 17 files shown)


services/one-stream/package/buildAnyPackage.js
```diff
@@ -256,12 +256,17 @@ export const buildAnyPackage = async (packageId, config) => {
...
-                  "yyyy-LL-dd"
-                )
-              : emptyField;
+          if (finalDispositionStatuses.includes(oneMacStatus)) {
+            putParams.Item.finalDispositionDate = DateTime.fromMillis(
+              anEvent.STATE_PLAN.STATUS_DATE
+            ).toFormat("yyyy-LL-dd");
+            // have to add 18000000 seconds because of time zone
+            putParams.Item.finalDispositionTimestamp =
+              anEvent.STATE_PLAN.STATUS_DATE + 18000009;
```
bflynn-cms, line 265: I see what you are doing here since the stored date from seatool is an epoch timestamp with no time component and you are therefore forcing it to US Eastern UTC -5.
However you exposed the same flaw has existed on the original implementation converting it to yyyy-LL-dd format. Due to this I am seeing different dates on the table vs the details page for final disposition date since details still uses finalDispositionDate and not finalDispositionTimestamp

bflynn-cms (changes requested):
See comment about timezones and package list vs details page. If you dont think the pre-existing issue exposed by your proper implementation needs to be dealt with now then thats fine but definitely wanted to point it out and do think it will be an issue.
But basically you being thorough and doing it right this time exposed the timezone flaw that was already there!

bflynn-cms (approved):
Nice job on the quick turn around!
I do like how both places are able to use the same value now.
Testing looks good and the dates are not "tainted" by my Central Timezone! :)
Approved.

PR #1417, opened 2023-12-04
VAL FIX: Fix NOSO building and the rebuild script needs to include Temporary Extensions

kristin-at-theta:
### Details
Testing in VAL uncovered a bug in the NOSO package building.
Also, the rebuild packages is dependent on there being a SEA Tool event, but Temporary Extension Requests do not have a corresponding SEA Tool event, so need to be put into the builder on their own.

### Changes
- fixed that subStatus handling required reverseChrono array to have at least one event, because NOSOs do not necessarily have any events.
- updated rebuildPackages to include the submittemporaryextension, submittemporaryextensionb, and submittemporaryextensionc events.
- brought in some test fixes from pi1-25598 branch

### Implementation Notes
- Doing this work in DEV first because the hotfix merge situation got really messy last time....

### Test Plan
1. Everything should build and test cleanly
2. Use the AWS lambda, createOneMacPackage
3. Verify there are no errors in the NOSO creation
4. Use the AWS Lambda, rebuildPackages (event requires a {"reason": "some reason"} format)
5. Verify that Temporary Extension packages are rebuilt.


bflynn-cms (approved):
approved!

PR #1423, opened 2023-12-08
PI2 26711: Can't take withdraw RAI action on another RAI in the same package

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-26711

NOTE: I named the branch for the other story... which we probably should not do at this time...

### Details
There was a bug in the enableRAIWithdraw function that was preventing the handling of multiple RAIs in a package.

### Changes
- changed the RAI Response sorting from String based to number based

### Test Plan
1. See story!


bflynn-cms (approved):
Approved!

PR #1424, opened 2023-12-08
Oy2 26709

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-26709

### Details

Change to withdraw rai status text

### Changes

- replaced all instances
- added mapping to ui map
- did map translation on lookup for detail and package view (same as we do for currentStatus)

### DIFF
--- a/services/app-api/form/disableRaiWithdraw.test.js
+++ b/services/app-api/form/disableRaiWithdraw.test.js
@@ -81,7 +81,7 @@ const GSIQueryReturn = {
       clockEndTimestamp: 1697123859496,
       componentId: "MD-22-2200-KG",
       componentType: "medicaidsparai",
-      currentStatus: "RAI Response Withdraw Enabled",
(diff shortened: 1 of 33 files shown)


kristin-at-theta (approved):
linked story is incorrect, but I figured it out :P 

things look correct, and I like that we changed it fundamentally rather than incrementally.  in VAL, need to migrate, but PROD should be fine with just the rebuild, right? :)

bflynn-cms:
> linked story is incorrect, but I figured it out :P
> 
> things look correct, and I like that we changed it fundamentally rather than incrementally. in VAL, need to migrate, but PROD should be fine with just the rebuild, right? :)

Ah yes, I switched the 0 and 7.....updated the link just in case someone else comes along.
And yes i included in this commit the migrate script which should run in val automatically, else ill just trigger it and since PROD doesnt run migrate script automatically we are good there.

PR #1431, opened 2023-12-12
PI2 26710: fix option 2 for -- Can't take withdraw RAI action after erroneous Withdraw Package action

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-26710

NOTE: when the CMS user enables the RAI Response Withdraw, the lambda is not waiting for the correct data change and shows the old state of the package.  Will fix this while the rest is getting reviewed....

### Details
One solution is to double up on the Withdraw Package action to also Disable any Enabled RAI Response Withdraws.  This means that CMS will have to re-enable the RAI Response Withdraw, but avoids the status-lock paradox.

### Changes
- refactor the disable lambda to allow another lambda to call the function
- call the disable lambda after the withdraw package action (for each package type)
- added seed-data to assist in testing

### Test Plan
1. Log in as a submitting user
2. Find one of (CHIP SPA, Medicaid SPA, Initial Waiver, Waiver Renewal, Waiver Amendment, Appendix K) with the RAI Response Withdraw enabled
3. Withdraw the package
4. Verify the package status is "Withdrawal Requested"
5. Verify the admin changes talk about the disable
6. Verify RAI response withdraw is disabled
7. Use a SEA Tool record to return the package status to "Pending" ("Under Review")
8. Login as a CMS user
9. Enable the RAI Response withdraw
10. login as Submitting user
11. submit an RAI Response withdraw
12. Verify the status is "Formal RAI Response Withdrawal Requested" (or something like that)

### DIFF
--- a/services/app-api/form/disableRaiWithdraw.js
+++ b/services/app-api/form/disableRaiWithdraw.js
@@ -20,7 +20,7 @@ export const disableRaiWithdrawFormConfig = {
     return userRole.isCMSUser;
   },
   appendToSchema: {
-    parentId: Joi.string().required(),
+    parentId: Joi.string(),
     parentType: Joi.string().required(),
     submissionTimestamp: Joi.date().timestamp(),
     adminChanges: Joi.array().items(


bflynn-cms (approved):
Everything looks good! I like the format/name chosen for showing admin changes of the withdraw event.
Testing everything panned out as well. Though I did have a moment of confusion when looking at data under the ticket number 26710  and could figure out why its not matching the actions I took........before realizing the branch name was different!! LOL but thats on me!
Approved!

kristin-at-theta:
heh heh - not totally on you - somehow I used 26710 as the branch name when I did the 26711 story, so I didn't have 26711 anymore as a possible branch name, so I figured I would do a full swap :P

PR #1433, opened 2023-12-12
Wthdrwrai

bflynn-cms:
Fix bug for withdraw RAI from packge list (after withdrawing another RAI from details page) where it wasnt redirecting back to package list post submit.

Testing Steps
1. Login as a state submitter
2. locate a package in Under Review with Withdraw Formal RAI Enabled
3. go to details page and withdraw package
4. go back to package list and locate another package in Under Review with Withdraw Formal RAI Enabled
5. withdraw package from the package list actions
6. verify you are redirected back to package list


kristin-at-theta (approved):
Hooray!  Great find :) 

Tests correctly!  Approved

PR #1439, opened 2024-01-02
PI2 26571: Setting Up Google Analytics (OneMAC)

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-26571

### Details
The complete task is to migrate the Google Analytics from OneMAC to MAKO - but GA in OneMAC does not appear to be working.  These updates should get the GA tagging set per deployment environment using the Action Secrets to hold the values for the different branches.

### Changes
- updated the tag code from Google because old tags/code wasn't working (GA was updated last summer)
- Added the REACT_APP_GOOGLE_TAG parameter to the GitHub Action secrets
- updated the index.html to use the stored value for the tag

### Implementation Notes
- must put REACT_APP_ in front of environment variables to get used in index.html ?? not entirely sure why, but it didn't work without it and does work with it... so.... 

### Test Plan
1. Create a new branch from this branch
2. Deploy branch
3. Verify that the application's Google Tag is the default "notag"
4. Create an Action Secret with <branch_name>_REACT_APP_GOOGLE_TAG and "newtag" (suggested value)
5. re-deploy (might not need to?  maybe can just restart the application...)
6. Verify that the application's Google Tag is new value


services/ui-src/public/env-config.js
```diff
@@ -13,4 +13,6 @@ window._env_ = {
   ALLOW_DEV_LOGIN: "true",
   METRICS_USERS: "placeholder",
   IS_OFFLINE: "true",
+  GOOGLE_TAG: "placeholder",
```
bflynn-cms, line 16: is this a leftover test? I dont see it being utilized.

bflynn-cms (approved):
Love how you moved the configuration out of code logic and into.....well.....configuration! :) (though I abhor gitlab oneway secrets)
Expanding on that, I do see the properties set in gitlab secrets for the prod, val, and dev  environments so have to assume they are correct so its nice you thought to preadd those.
Test branch seems to pull from secrets and updates when the secret is updated and app is redeployed.
While I dont know enough about google analytics to verify everything, it does seem to be doing what was intended.
Approved!

services/ui-src/public/env-config.js
kristin-at-theta: Honestly, I'm not sure if those env variables ever get used, I thought they would be for local.... so I gave them different values to see if it mattered and well, forgot to add the pretext "REACT_APP_" to that one :P 

Thanks for the catch!

PR #1443, opened 2024-01-23
PI2 27030: Retry Mechanism for Failed Emails

kristin-at-theta:
Story: https://qmacbis.atlassian.net/browse/OY2-27030
Endpoint: https://d25qi725grz9hh.cloudfront.net/

### Details
Add an email capture to OneMAC so that failed emails can be resent.  Though.... AWS SES automatically retries both soft bounces and delivery delays, which are the two cases that are the best choices for resending... hard bounces, etc. need additional triage on why the email sending failed and should not be resent as-is.

### Changes
- added email event SNS topic for use by AWS SES
- created a configuration set that should be used by all emails from one branch
- added the email service
- added processEmailEvents to update the progress of each email
- updated submitAny and the user change events to save the emails

### Implementation Notes
- It made sense to save the emails in DynamoDB for OneMAC Live, but this is not the solution being created for MAKO.

### Test Plan
1. Login as a user for whom you have email access to the emails sent.
2. Perform an event that sends an email
3. Access the AWS console to verify that the email is saved and that email events are being captured.


services/app-api/utils/scheduleEmail.js
bflynn-cms, line 1: ooops!

bflynn-cms (approved):
I do like how this is saving the emails in dynamo and despite previous conversations with platform team I think it would useful to have these records for debugging/proof.

Approved

services/email/handlers/processEmails.js
```diff
@@ -0,0 +1,60 @@
...
+    TableName: oneMacTableName,
+    IndexName: "GSI2",
+    KeyConditionExpression: "GSI2pk = :pk",
+    ExpressionAttributeValues: {
+      ":pk": "Send",
+    },
+  };
+  try {
+    const result = await dynamoDb.query(queryParams).promise();
+    console.log("results are: ", result);
```
bflynn-cms, line 24: Is this just querying but not doing anything? I guess im not following the entire workflow

services/app-api/utils/scheduleEmail.js
kristin-at-theta: deleted!

services/email/handlers/processEmails.js
kristin-at-theta: So... at this time, yes, processEmails queries the database to show how to get the emails for a given status, BUT, I have no idea which statuses they want to do anything with and what to do with them.

The story is to "retry failed messages," BUT, the soft failures that we would conceivably want to retry (as in, it's not a bad email address and will always be bad) have actually already been retried by AWS SES several times before the final reporting.  So.... story AC is already met, but I thought it might be good to leave in the "how" to discover a truly failed email.

PR #1476, opened 2024-06-20
Files section in Submission emails for all package types shows wrong file name type

andieswift:
Story: https://qmacbis.atlassian.net/browse/OY2-28880

### Details

### Changes

- added the Mako attachment hash map to the email package detail page

### Implementation Notes

- just added this in the specific code block needed, not sure if we want to abstract this map to use across the app?

### Test Plan

1. Test step 1
2. Test step 2


andieswift:
Wrong Repo 😅

PR #1494, opened 2024-08-13
oy2-29115 - update package builder to set inactivated status on package

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-29115

### Details

Package builder was properly setting the package to not display by not setting a GSI key; however the inactivated status was not carrying over.

### Changes

- Updated package builder to track if any OneMac record is marked as currentStatus = Inactivated and if override the package status

### Test Plan

1. Identify a package
2. Run softDeleteComponent lambda
https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/admin-bcf-29115-softDeleteComponent?tab=testing
3. Verify Package record in DB has updated status and that the package no longer appears in the dashboard

### DIFF
--- a/services/one-stream/package/buildAnyPackage.js
+++ b/services/one-stream/package/buildAnyPackage.js
@@ -80,6 +80,7 @@ export const buildAnyPackage = async (packageId, config) => {
     let showPackageOnDashboard = false;
     let lmTimestamp = 0;
     let adminChanges = [];
+    let packageInactivated = false;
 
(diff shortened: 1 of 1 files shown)


andieswift (approved):
Tested the soft delete of a new package I created: MD-24-1241-ANDI and the status was changed in the db. The package was also removed in the UI dashboard. 
Code makes sense and is clean awesome work! :)

PR #1506, opened 2024-09-17
Oy2 29781 2

anyoussefinia:
Story: https://qmacbis.atlassian.net/browse/OY2-29781

Endpoint: (https://d3patzhc2mhn3y.cloudfront.net/)

### Details

Remove MMDL related cards, routes and pages


andieswift (approved):
I think the code looks good. One minor thing, I think you should also remove the property MMDL from 'ExternalSystem' in TriangeExternalLandingPage.tsx. You could change that whole value from an object to a const if you wanted to, but I'll leave it up to you. 

But that is super minor so I am going to mark this as approved. Also I agree with you I think the language on the chip spa selection should change. I see your comment on Jira but you might want to add a screenshot and at Hannah & LaTisha

PR #1525, opened 2024-10-03
Oy2 30157

anyoussefinia:
Story: https://qmacbis.atlassian.net/jira/software/c/projects/OY2/boards/181?selectedIssue=OY2-30157
Endpoint: https://d1psn7u3jo8wds.cloudfront.net/


services/ui-src/src/libs/formLib.tsx
```diff
@@ -181,11 +194,10 @@ export const defaultConfirmSubmitMessageSubsequentSubmission = (
 );
 
 export const defaultConfirmSubsequentSubmission: ConfirmSubmitType = {
-  confirmSubmitHeading: "OneMAC only for document submission",
+  confirmSubmitHeading: "Submit additional documents?",
   confirmSubmitMessage: (
     <p>
-      States and CMS reviewers will communicate about the submission through
-      email.
+      These documents will be added to the package and reviewed by CMS.
```
andieswift, line 200: This doesn't seem to be part of the ticket I may be missing something but why did you make this change?

anyoussefinia: yes it was picked up by a chang from a different ticket that we merged into master/masterclone. Will open up a separate PR to dev for this change.

andieswift (approved):
Awesome!

PR #1528, opened 2024-10-07
Masterclone

anyoussefinia:
Story: https://qmacbis.atlassian.net/browse/OY2-30490
Endpoint: https://d38apkewa2hawd.cloudfront.net/


andieswift (approved):
If Valencia already approved this on masterclone then I think this is good 💛

PR #1536, opened 2024-10-17
Oy2 30326 2

anyoussefinia:
Story: https://qmacbis.atlassian.net/jira/software/c/projects/OY2/boards/181?selectedIssue=OY2-30326
Endpoint: https://dn2ot3bby1tqp.cloudfront.net/

this will remove any version 4.17.01 and above i.e. 4.17.99, 4.18.01, 5.01.01, 9.01.01. this will hopefully cover any loadsh major/minor upgrades

Lodash Version does not appear: 
Screenshot 2024-10-17 at 1 52 26 PM

### DIFF
--- a/services/ui-src/remove-versions.js
+++ b/services/ui-src/remove-versions.js
@@ -5,7 +5,7 @@ const path = require('path');
 const buildDir = path.join(__dirname, 'build', 'static', 'js');
 
 // Define the regular expression to match the version pattern
-// Specifically targeting the line with Bn.VERSION
+// specifically targeting the line with Bn.VERSION
 const regex = /(Bn\.VERSION = ")(4\.(1[7-9]|[2-9][0-9])\.\d{2}|5\.\d{2}\.\d{2}|[6-9]\.\d{2}\.\d{2})(")/g;
 
 // Function to remove the version from files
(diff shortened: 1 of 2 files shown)


services/ui-src/remove-versions.js
```diff
@@ -0,0 +1,38 @@
...
+    files.forEach(file => {
+      const filePath = path.join(dir, file);
+
+      // Check if the file is a JavaScript file
+      if (file.endsWith('.js')) {
+        fs.readFile(filePath, 'utf8', (err, data) => {
+          if (err) throw err;
+
+          // Replace the version pattern
+          const result = data.replace(regex, ''); // Remove matched versions
```
bflynn-cms, line 25: while this appears to work, I am weary about replacing based just on any version number in any file
maybe its fine but worried it could overstep and erase things outside the scope of this change.
Any reason we couldnt search for the specific string that proceeds the package we are trying to obfuscate?

anyoussefinia: it looks like the specific string that proceeds the package is "Bn.VERSION = " which appears only once in the file so I think this should actually work. will see if I can get something like that working

bflynn-cms (approved):
Thanks for making those changes! Looks good!

Dark-Knight-1313 (approved):

Dark-Knight-1313 (approved):

PR #1545, opened 2024-11-13
Oy2 30766 0

anyoussefinia:
Story: https://jiraent.cms.gov/browse/OY2-30764
Endpoint: https://d1m9a2b5yio8n6.cloudfront.net/

### Details

This Story has to do with using launch darkly flags to toggle mmdl Banner/Notification and FAQ PDF's and Implementation guides.

### Changes

added launch darkly client as a wrapper component to the application, and two flags "mmdlNotificaitons" & "mmdlFAQ" to the Home.js and App.tsx pages. 

### Implementation Notes

Toggle the flags using the launch darkly console accessed through mini orange: https://mo-idp.cms.gov/enduserwelcome

### DIFF
--- a/services/ui-src/src/App.tsx
+++ b/services/ui-src/src/App.tsx
@@ -192,9 +192,9 @@ const  App = () => {
   }, [mmdlNotification]);
 
   useEffect(()=>{
-    console.log("use effect #2");
-    console.log("notification State:  " +notificationState);
-    console.log("authstate.isAuthenticated:" + authState.isAuthenticated);
+    // console.log("use effect #2");
+    // console.log("notification State:  " +notificationState);
+    // console.log("authstate.isAuthenticated:" + authState.isAuthenticated);
     (async ()=> {
       try{
         if(notificationState && authState.isAuthenticated) {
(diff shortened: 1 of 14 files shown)


services/ui-src/src/App.tsx
```diff
@@ -171,6 +151,68 @@ export function App() {
     setUserInfo();
   }, [setUserInfo]);
 
+  useEffect(() => {
+    if (mmdlNotification !== undefined) {  // Ensure the flag has been resolved
+      setNotificationState(mmdlNotification);
+    }
+  }, [mmdlNotification]);
```
andieswift, line 158: Having this part of the code be in its own useEffect seems like extra work/slows down the app

How about instead, move this line of code 
`const {mmdlNotification} = useFlags();` into the useEffect on line 160, and get rid of the dependencies in the array at the end of the useEffect. that way this will always run on the start of the app

andieswift (changes requested):
I think this looks good, I added one comment that I think should be addressed and do you think you could add a video recording showing turning on and off the flags and seeing how that effects the frontend?

Also thanks for adding those tests!! those look great imo

anyoussefinia:
[Uploading screen-capture (2).webm…]()

anyoussefinia:
[screen-capture (2).webm](https://github.com/user-attachments/assets/3ab22235-bb3e-4e6a-8d48-07b23ce36b5b)

anyoussefinia:
[screen-capture.webm](https://github.com/user-attachments/assets/ae198952-b21e-4ba3-bad1-803c0288479c)

andieswift (approved):
Thanks for the video, I think this looks good. We might want to refactor those useEffects
 in the future but since we are on a time crunch this is good for now

PR #1552, opened 2024-11-29
Launch Darkly for Subsequent Documentation

bflynn-cms:
Story: https://qmacbis.atlassian.net/browse/OY2-31108

### Details

Wrap subsequent submission functionality in launch darkly flags

### Changes

- Added server side launch darkly capabilities
- changed function for getting actions to be async in order to take advantage of LD async calls
- created filter for routes in order to utilize LD flags

### Implementation Notes

- Working on moving LD server side functionality into a Lambda Layer for code reuse so that it can be used across the application


anyoussefinia (approved):
Looks Good! does this mean we can toggle on Sub Sub for various roles instead of all users at once?

bflynn-cms:
> Looks Good! does this mean we can toggle on Sub Sub for various roles instead of all users at once?

No, that would be cool though. I think Launch Darkly does support that if you pass it certain data about each user. I only changed the role processing logic in order to move it to handle the async call which the .map method couldnt handlel.

PR #1555, opened 2024-12-05
Pentest

anyoussefinia:
https://jiraent.cms.gov/browse/OY2-31462
 https://jiraent.cms.gov/browse/OY2-31468
 https://jiraent.cms.gov/browse/OY2-31474

Endpoint:  https://d323ybbxjp59gr.cloudfront.net/

### Details

Added an attribute to the id JWT that contains user roles & checks those roles in order to access UserManagement (/usermanagement) route. 

Also removed the ability to call UserDataApi with email as a query parameter from the UserPage (/profile) route

### Test Plan

1. We can test that user cannot make an api call from the /profile page by adding the email to the url (See Video)

2. Since the pentest results found the vulnerability by modifying the response to the authentication request, this will need to be retested by them


anyoussefinia:
pentester found that user can make an api call from the /profile page by adding the email to the url. This was removed. 
[screen-capture (14).webm](https://github.com/user-attachments/assets/614a5f28-390c-4770-b55d-b2efe292e8a0)

services/ui-src/src/Routes.tsx
```diff
@@ -155,11 +205,11 @@ const ROUTE_LIST: RouteSpec[] = [
...
-      {
-        path: ROUTES.PROFILE + "/:userId",
-        exact: true,
-        component: UserPage,
-      },
+      // {
+      //   path: ROUTES.PROFILE + "/:userId",
+      //   exact: true,
+      //   component: UserPage,
+      // },
```
bflynn-cms, line 212: We still need this route because admin users should be able to see user profile pages from clicking the link on the User Management page. We just need to inject code to check the token to ensure they have the proper role in the jwt to access this route. Maybe similar to how you locked the User Management page but it would be more complex because we would want to use whatever rules are used to populate the user management page to ensure its only a certain group of users as well. AKA Admin doesnt mean you can see everyone. For example state admins should be be able to see user profiles of users in a state for which they are a state admin.

bflynn-cms (changes requested):

PR #1566, opened 2024-12-13
Pentest

anyoussefinia:
Story: https://jiraent.cms.gov/browse/OY2-31468
https://jiraent.cms.gov/browse/OY2-31474
Endpoint: https://d108315ije1eu7.cloudfront.net/

### Details

added a batch function that modifies the idToken JWT with a "custom:user_roles" attribute with the roles retrieved for each user from dynamo. 

Added route protection for the /usermanagement route which checks the jwt for proper role access

### Changes

-also added the trigger function for potential use if the intermittent functionality problem is fixed. 

### Test Plan

Just a smoke test to make sure nothing is broken, The pen testers will need to verify that the usermanagent screen cannot be accessed by modifying the response to UserDataApi.userProfile by setting the admin role status from "revoked" to "active"


services/ui-src/src/Routes.tsx
```diff
@@ -139,17 +141,73 @@ const SignupGuardRouteListRenderer: FC<{ routes: RouteSpec[] }> = ({
...
+    idTokenKey && localStorage.getItem(idTokenKey[0]);
+    if (!idToken) return false;
+    const decodedIdToken: any = jwt_decode(idToken);
+     userRoles = decodedIdToken["custom:user_roles"];
+  } catch (error) {
+    console.error("error decoding idToken", error);
+    return false; 
+  }
+
+  const allowedRoles = [
```
bflynn-cms, line 167: thanks for turning this into a whitelist of allowed roles!

services/ui-src/src/containers/UserManagement.js
```diff
@@ -106,6 +106,7 @@ const UserManagement = () => {
         console.log("Error while fetching user list.", error);
         setAlertCode(RESPONSE_CODE[error.message]);
       });
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
bflynn-cms, line 109: please discard this change as no functional changes were made to this file

services/ui-src/src/containers/UserPage.js
```diff
@@ -161,7 +161,7 @@ const UserPage = () => {
   const { userProfile, setUserInfo, updatePhoneNumber, userRole, userStatus } =
     useAppContext();
   const location = useLocation();
-  const { userId } = useParams() ?? {};
+  let { userId } = useParams() ?? {};
```
bflynn-cms, line 164: curious why this change was made........might even be flagged by lint if setting with let but is never overridden.....but also probably userId is something we dont want to let get overwritten?

anyoussefinia: you're right I did this when I was considering how to encrypt the email/userId before Andy fixed it

bflynn-cms (approved):

PR #1570, opened 2024-12-30
feat(banner): Banner for Subsequent Document Upload Release

andieswift:
Story: https://jiraent.cms.gov/browse/OY2-31867
Endpoint: https://d2s15t18jolyzj.cloudfront.net/
### Details

Create a banner in OneMAC so that when either a State or CMS user logs into the OneMAC system, the banner will appear to have nomenclature related to the new Subsequent Document Release. 

### Changes

- removed launch darkly flag wrapper over the WHOLE notifications
- Now NotificationAPI will always be called and _then_ we will check if we hide/show a notification

### Implementation Notes

- I will have to run the lamda on each respective branch to see the new subdoc notification
- I think we could make this code more bullet proof going forward. What if we add a new property to the notification type for the flags then on the frontend use that property to check if the notification shows/hides

Screenshot 2025-01-02 at 11 52 12 AM

### Test Plan

1. Log in as any user, check the home page has the notification
2. check the user notification is displayed
3. Try again with the `enableSubsequentDocumentation` flag turned off and check that the notification is NOT there


Dark-Knight-1313 (approved):
Hi Andi. looks good. I have one quick question. When you get a chance, ping me.

Also, I love you idea ( see below )

I think we could make this code more bullet proof going forward. What if we add a new property to the notification type for the flags then on the frontend use that property to check if the notification shows/hides

PR #1577, opened 2025-01-13
input validation matching phone number format

anyoussefinia:
Story: https://jiraent.cms.gov/browse/OY2-32232
Endpoint: https://d1ymop2ngbtmlb.cloudfront.net/

### Details

Input validation for phone numbers

### Changes

We decided on only allowing numbers and programmatically adding () and - as the user types

### Test Plan

Simply try special characters while typing a phone number and see that it does not allow


Dark-Knight-1313 (approved):
Looks good

PR #1586, opened 2025-02-06
Pentest 31462

anyoussefinia:
Story: https://jiraent.cms.gov/browse/OY2-32701
Endpoint: https://d1qcvu24747qjj.cloudfront.net/

### Details

admin users can visit the /profile/encoded-email page for any user. non admin users can only view other users with matching territories. 

### Changes

Pentesters identified that any user can visit the /profile/encoded-email page for any other user. Fix to block users that don't have permissions.

### Test Plan

1. Log in as an admin user and see that you can visit any other user page regardless of matching territores.
2. Log in as non-admin user and verify that you can see another non-admin user with some matching territory.
3. verify that non-admin user cannot view other users (admin or non admin) that do not have matching territories.


anyoussefinia:
[Uploading screen-capture (16).webm…]()

anyoussefinia:
[screen-capture (16).webm](https://github.com/user-attachments/assets/6d4050ee-87e2-4d35-9820-a463cefb1cef)

andieswift (changes requested):
`getUserV2` is a confusing name, do you think you could either rename this file to match more of what you are doing or make this work part of getUser.js? 

Also could you provide me with a user outside of MD so I can test if I can view them? I tried looking with the admin login but I still only see MD packages.

andieswift (approved):
I like this file name a lot more thank you, I think this is good 👍

PR #1601, opened 2025-03-18
Oy2 33622 fix for role updates

anyoussefinia:
Story: https://jiraent.cms.gov/browse/OY2-33622
Endpoint: https://d3aofzk1iaqs57.cloudfront.net/

### Details

A PR to fix admin user inability to approve/revoke/deny roles 

### Changes

-added dynamo get call to check if latest is non numeric 
-modify set statement for update params
-added extra logging for update failures 

### Test Plan

1.  Ensure roles can be approved/denited/revoked


anyoussefinia:
[screen-capture (22).webm](https://github.com/user-attachments/assets/51d1209d-e530-483f-8955-891aced30328)

andieswift (approved):
looks good,
minor thing you added/removed spaces in updateUserStatus, and git is highlighting changes that don't look like changes it makes me thing you might need to update your formater to match the rest of the repo; but its not support important I feel

services/app-api/utils/changeUserStatus.js
```diff
@@ -24,57 +24,79 @@ export const changeUserStatus = async ({
...
+ // add a new v0 and v latest
+  const updateParams = {
+    TableName: process.env.oneMacTableName,
+    ReturnValues: "UPDATED_NEW",
+    Key: {
+      pk: email.toLowerCase(),
+      sk: `v0#${role}#${territory}`,
+    },
+    UpdateExpression:
+      "ADD Latest :incrval SET " +
```
bflynn-cms, line 36: Im not sure this is a better implementation than the original handling using if_not_exists. Both implementations seem to be functionally identical so not understanding why the change was needed. Also AWS documentation suggests to use SET over ADD operation: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.ADD

services/app-api/utils/changeUserStatus.js
```diff
@@ -24,57 +24,79 @@ export const changeUserStatus = async ({
...
-        ":GSI1sk": buildSK(role, territory),
-        ":GSI2pk": `${role}#${territory}`,
-        ":GSI2sk": status,
-        ":defaultval": 0,
-        ":incrval": 1,
-      },
-    };
+    // fix for latest being null or undefined
+    const result = await dynamoDb.get(getParams);
+    if (typeof result.Item?.Latest !== "number") {
```
bflynn-cms, line 82: Again im not following the changes here because the original implementation used if_not_exists which would cover null and undefined use cases.

bflynn-cms (commented):
Left some comments for clarification on changes.

bflynn-cms (changes requested):
Please check my comments. Thanks!

services/app-api/utils/changeUserStatus.js
bflynn-cms: ok relooking i do see that this use case would also cover invalid character data that is not numerical and not just null or undefined

bflynn-cms (approved):

PR #1604, opened 2025-04-29
Modify the 90 day clock expiration on the chip spa rai response email 

anyoussefinia:
Story: https://jiraent.cms.gov/browse/OY2-34209
Endpoint: https://d3ihudd5bupj3j.cloudfront.net/

### Details

Modify the 90 day clock expiration on the chip spa rai response email 

### Changes

calculate the clock end based on various timestamp info instead of simply saying 90 days from the day the rai response is submitted


andieswift (commented):
I see how you are getting the 90 days but I am confused as to how that will update in the emails? won't the work here be overwritten by the code in stateSubmissionReceipt on line 11?

Or I am I missing something?

anyoussefinia:
> I see how you are getting the 90 days but I am confused as to how that will update in the emails? won't the work here be overwritten by the code in stateSubmissionReceipt on line 11?
> 
> Or I am I missing something?



> I see how you are getting the 90 days but I am confused as to how that will update in the emails? won't the work here be overwritten by the code in stateSubmissionReceipt on line 11?
> 
> Or I am I missing something?

Basically line eleven just formats any timestamp i.e. 1745510735395 will change to something like Tuesday, August 5th.... 

The main point is to set the clock end timestamp on submitAny line 157 then use that in stateSubmissionReceipt.

andieswift (approved):

